PR #19, opened 2022-07-14
Fix Model Classes

### LINKED ISSUE #18: Preprocess OpenAPI JSON
The issue with the SwaggerCodegen is that it cannot handle "oneOf, anyOf, allOf" of OpenAPI 3.0. As a result, the generated code usually isn't functional. We have researched on other Java code generator and the result is similarly disappointing.

We handled it by isolating the scope of the generated code, basically only the model classes and Retrofit methods. This greatly reduced our work to work with the generated code. However, the model classes still needs some manual fix to rename and sometimes replace the "oneOf, anyOf, allOf" in the class name. This doesn't sound like a big deal until we realized that we had more than 90 of model classes that potentially need manual fix.

Other than waiting for the OpenAPI generators to fix their bugs, we could something on our end: remove the use of "oneOf, anyOf, allOf" in the output swagger.json from APIServer's Swagger page. Here is the plan:

1. Understand OpenAPI 3.0 and the semantic meaning of the "oneOf, anyOf, allOf" keywords.
2. Check out swagger.json and see if we can find a way to preprocess it and get rid of the above keywords.
3. Test that the processed output can work by diffing its model class with the existing classes.

lf-jason-jia:
Fix #18 

As a result, we found the previously generated model classes have mistakenly generated some classes with type "Object" whereas the more specific class exists. This PR uses the preprocess swagger.json to produce the correct model classes.

1. Generate correct model classes using preprocess JSON.
2. Optimized imports for all classes.
3. Removed some useless comments from generated code (we could keep them in the future we want minimal human intervention).


steven-xu-lf (approved):

src/main/java/com/laserfiche/repository/api/clients/impl/model/AcceptedOperation.java
```diff
@@ -1,15 +1,3 @@
-/*
```
steven-xu-lf, line 1: just wondering, should we also remove the corresponding comments for both the C# and js client libs?

lf-jason-jia: I guess either way is fine. I did this without knowing that we are going to the pipeline that goes from retrieval of the swagger.json to a fully working project. If we go that path, I guess we probably need a custom script to delete those comments which I don't think is easy to write (it might break the source file). So, maybe we can keep that in the future?

lf-jason-jia: Sorry for delay on this PR. I am still working on the Maven issue. BTW, if you are able to run the tests just fine, I guess we can merge it (I don't want my issue delays the PR).

PR #20, opened 2022-07-14
Add test header and branch protection

steven-xu-lf:
https://tfs/DefaultCollection/Server%20Tools/_workitems/edit/386866
issue -> https://github.com/Laserfiche/lf-repository-api-client-java/issues/21


src/test/java/integration/BaseTest.java
```diff
@@ -75,4 +77,10 @@ public static CompletableFuture<Boolean> allFalse(List<TemplateFieldInfo> arr) {
         }
         return CompletableFuture.supplyAsync(() -> true);
     }
+
+    private static RepositoryApiClientImpl createTestRepoClient(String servicePrincipalKey, AccessKey accessKey, String testHeader) {
+        RepositoryApiClientImpl testRepoClient = RepositoryApiClientImpl.CreateFromAccessKey(servicePrincipalKey, accessKey);
+        testRepoClient.okBuilder.addInterceptor(chain -> chain.proceed(chain.request().newBuilder().addHeader(testHeader, "true").build()));
```
jason-chan-lf: Is there a way to add default headers to the Repository Client in an easier way, like for the js and c# clients? Can be added in a different PR if that is possible. This way you wouldn't need to make the okBuilder public

steven-xu-lf: from my understanding, the only way to add/set headers that will be passed through for every request for retrofit (i.e. authorization and loadtest) is to add them in the Oauth interceptor. Right now, the only alternative solutions I can think of involve creating and instantiating additional objects specifically used for the tests or adding some functions for testing in the RepositoryAPIClientImpl class

steven-xu-lf: oh wait, i think their might be a specific java package that will enable me to access private variables outside class, looking into this a little closer......

steven-xu-lf: ok so 2 shorter alternative solutions that i am thinking of here that don't involve creating an additional test object

1. creating and using getters/setters in the repositoryApiClientimpl object 
2. using the java.lang.reflect package to get the private variable (currently looking into that)

jason-chan-lf: I feel like it would be a bit cleaner if there was a function in `RepositoryApiClientImpl` for setting default headers. Like it could be `RepositoryApiClientImpl.SetDefaultRequestHeaders(Dictionary headers)` and inside the function, it would set the headers on the private okBuilder

steven-xu-lf: ok added a `setDefaultHeaders` and a `setLoadTestHeader` function in `RepositoryApiClientImpl`

src/main/java/com/laserfiche/repository/api/RepositoryApiClient.java
```diff
@@ -55,4 +55,6 @@ public interface RepositoryApiClient {
      * @return The Laserfiche Repository Template Definitions API client.
      */
     TemplateDefinitionsClient getTemplateDefinitionClient();
+
+    void setLoadTestHeader(String testHeader);
```
lf-jason-jia: This is for test only and it seems like we could hide it from the public.

lf-jason-jia (changes requested):
See comments

src/main/java/com/laserfiche/repository/api/RepositoryApiClient.java
lf-jason-jia: Like @jason-chan-lf said in another comments, we could add that method to `RepositoryApiClientImpl` instead of the public interface.

steven-xu-lf: ok removed it from the interface

pom.xml
```diff
@@ -1,8 +1,8 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
-  <groupId>io.swagger</groupId>
-  <artifactId>lf-repository-api-client-java</artifactId>
+  <groupId>com.laserfiche</groupId>
+  <artifactId>com.repository.api.client</artifactId>
```
lf-jason-jia: Using dash is the convention. Please revert the change.

lf-jason-jia: Also, we probably need to change it for client core lib.

pom.xml
```diff
@@ -1,8 +1,8 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
-  <groupId>io.swagger</groupId>
-  <artifactId>lf-repository-api-client-java</artifactId>
+  <groupId>com.laserfiche</groupId>
```
lf-jason-jia, line 4: Thanks for catching it!

lf-jason-jia:
Also, even though you could link the issue with the PR, it would be nice if in the future we use branch name in the form of issue/{issue_#}-some-optional-words

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -57,6 +54,15 @@ private <C> C createClient(Class<C> clientInterface) {
                 .create(clientInterface);
     }
 
+    private void setDefaultHeaders(String servicePrincipalKey, AccessKey accessKey) {
```
lf-jason-jia: This method is called `setDefaultHeader` but its content doesn't really set the header. Maybe a better name?

lf-jason-jia: I think overall the `setLoadTestHeader` and `setDefaultHeaders` feel a bit out of place for me. Can you give me some time to see if there's a better alternative?

pom.xml
steven-xu-lf: the groupID i think has already been registered on sonatype so i guess i will just replace the periods with dashes in the artifact id

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
steven-xu-lf: how about `setDefaultHeaderInterceptor` or `setLoadTestHeaderInterceptor`?

pom.xml
lf-jason-jia: Hi @sxu1998 , Do you mean the one above this line? I think you don't need to register artifact ID?

steven-xu-lf: yes, the group ID has already been registered and yes i don't believe the artifact ID needs to be registered in sonatype

src/test/java/integration/BaseTest.java
```diff
@@ -14,21 +14,22 @@
...
+    protected static String testHeader;
     protected static RepositoryApiClient repositoryApiClient;
 
     @BeforeAll
     public static void setUp() {
         spKey = System.getenv("SERVICE_PRINCIPAL_KEY");
         String accessKeyBase64 = System.getenv("ACCESS_KEY");
         repoId = System.getenv("REPOSITORY_ID");
+        testHeader = System.getenv("TEST_HEADER");
         if (spKey == null && accessKeyBase64 == null && repoId == null) {
```
lf-jason-jia: Should probably also add the test header in the if-condition.

lf-jason-jia: Done

lf-jason-jia (dismissed):

src/test/java/integration/BaseTest.java
```diff
@@ -47,7 +51,12 @@ public static void setUp() {
         accessKeyStr = accessKeyStr.replace("\\\"", "\"");
 
         accessKey = gson.fromJson(accessKeyStr, AccessKey.class);
+
+        testHeaders = new HeadersImpl();
+        testHeaders.append("TEST_HEADER", testHeaderValue);
```
jason-chan-lf: should be `testHeaders.append(testHeaderValue, true);`

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -51,6 +50,15 @@ private <C> C createClient(Class<C> clientInterface) {
                 .create(clientInterface);
     }
 
+    public void setDefaultRequestHeaders(Headers defaultHeaders) {
```
jason-chan-lf: with this implementation, is it possible to overwrite/remove the default headers if this is called multiple times? Maybe we need to have Headers as a property in this class

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -51,6 +54,27 @@ private <C> C createClient(Class<C> clientInterface) {
                 .create(clientInterface);
     }
 
+    @Override
+    public void setDefaultRequestHeaders(Headers defaultRequestHeaders) {
+        this.defaultHeaders = defaultRequestHeaders;
+        okBuilder.addInterceptor(chain -> {
```
jason-chan-lf, line 60: If `setDefaultRequestHeaders` is called 3 times, will it add 3 Interceptors on the okBuilder in total? And inside the interceptor, it is adding the `defaultRequestHeaders` instead of `this.defaultHeaders`, so if `setDefaultRequestHeaders` is called 3 times, will all headers from these calls be added to the request, instead of overwriting previous default request headers?

Would it make sense to add this interceptor to the constructor of `RepositoryApiClientImpl` so that it only happens once?

jason-chan-lf (approved):

PR #23, opened 2022-07-21
Issue/22 refactor

### LINKED ISSUE #22: Refactor to make working with auto generated code easier
1. See this for the current [structure](https://github.com/Laserfiche/lf-repository-api-client-java/blob/1.x/DEVELOPER.md)
2. We want to do the following:
        1. Automatically retrieve swagger.json
        2. Preprocess the JSON
        3. Generate code from that JSON
        4. Only take the code we need from the JSON and have it nicely added to existing code.
        5. Delete the rest of the auto generated code
3. 2.4 isn't possible with current file structure (see 1). To do that, we will need to have some kind of separation between the auto generated code and manual code: we need to separate them into two files. When we run the automation in 2, we just override the auto generated code.

lf-jason-jia:
_I am sorry for the long PR, I wasn't able to reduce it to something smaller._

1. In package `com.laserfiche.repository.api.clients.impl`, each retrofit service interface either contains generated code (named xxxApi) or contains manual code (nextLink support or any manual override, named `xxxApiEx`). This way, we can have pipeline that swaps out generated code if our APIs change.
2. JavaDoc only stays in one place (single source of truth): if a method is implementing an interface, the JavaDoc stays with the interface; otherwise it stays with the method.
3. Move client instantiation logic into each of the clients themselves instead of leaking into `RepositoryApiClientImpl`.
4. Methods use and return public interface as much as possible instead of the corresponding implementation.

Fix #22

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -1,3 +1,0 @@
-<manifest package="com.laserfiche.repository.api" xmlns:android="http://schemas.android.com/apk/res/android">
-    <application />
-</manifest>
--- a/src/main/java/com/laserfiche/repository/api/BaseClient.java
+++ b/src/main/java/com/laserfiche/repository/api/BaseClient.java
@@ -17,7 +17,7 @@ public class BaseClient<T, S> {
     protected void setupClients(Class<T> generatedClientClass, Class<S> extensionClientClass) {
         this.generatedClient = createClient(generatedClientClass);
         // Not all API paths have extension client (meaning, they don't have custom methods such as nextLink or forEach)
-        if (extensionClient != null) {
+        if (extensionClientClass != null) {
             this.extensionClient = createClient(extensionClientClass);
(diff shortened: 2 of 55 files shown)


steven-xu-lf (commented):
it appears that all the foreach/next link test cases are failing, and i see that you added a bunch of additional *ex.java files which contains the paginate functions, so should all those test cases also be refactored?

lf-jason-jia:
> it appears that all the foreach/next link test cases are failing, and i see that you added a bunch of additional *ex.java files, so should all those test cases also be refactored?

Yes, you are right. I am working on that.

lf-jason-jia:
> it appears that all the foreach/next link test cases are failing, and i see that you added a bunch of additional *ex.java files which contains the paginate functions, so should all those test cases also be refactored?

Fixed, thanks!

steven-xu-lf (approved):

PR #32, opened 2022-09-13
Reimplement using Unirest and our own codegen

lf-jason-jia:
1. Use Unirest and our own codegen to generate most of the client code.
2. Add integration tests that have basic coverage.
3. Minor changes to the interface (mostly because the original API wasn't correct).


src/main/java/com/laserfiche/repository/api/clients/impl/SimpleSearchesClientImpl.java
```diff
@@ -0,0 +1,64 @@
...
+                    if (httpResponse.getStatus() == 401) {
+                        throw new RuntimeException("Access token is invalid or expired.");
+                    }
+                    if (httpResponse.getStatus() == 403) {
+                        throw new RuntimeException("Access denied for the operation.");
+                    }
+                    if (httpResponse.getStatus() == 429) {
+                        throw new RuntimeException("Operation limit or request limit reached.");
+                    }
+                    return httpResponse.getBody();
```
jason-chan-lf, line 61: Will this handle unexpected response codes correctly like 500? Would it be better to check the success status code to know to return the body?

src/main/java/com/laserfiche/repository/api/clients/impl/model/PostEntryChildrenRequest.java
```diff
@@ -1,142 +1,21 @@
...
-    this.entryType = entryType;
-  }
-
-  public PostEntryChildrenRequest targetId(Integer targetId) {
-    this.targetId = targetId;
-    return this;
-  }
-
-   /**
-   * The TargetId is only needed for creating a shortcut. This will be the entry ID of the shortcut target.
```
jason-chan-lf, line 66: should the generator add the javadocs to these model properties?

src/main/java/com/laserfiche/repository/api/clients/impl/model/PostEntryChildrenRequest.java
```diff
@@ -1,142 +1,21 @@
...
-   /**
-   * The type of the entry.
-   * @return entryType
-  **/
-  @Schema(description = "The type of the entry.")
-  public PostEntryChildrenEntryType getEntryType() {
-    return entryType;
-  }
-
-  public void setEntryType(PostEntryChildrenEntryType entryType) {
```
jason-chan-lf, line 56: is java coding conventions to use getters and setters? I noticed the model classes dont have this anymore

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -2,43 +2,47 @@
...
+    private final TagDefinitionsClient tagDefinitionsClient;
+    private final TasksClient tasksClient;
+    private final TemplateDefinitionsClient templateDefinitionsClient;
 
     protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKey, String baseUrlDebug) {
-        String baseUrl = baseUrlDebug != null ? baseUrlDebug : "https://api." + accessKey.getDomain() + "/repository/";
-        RepositoryApiDeserializer json = new RepositoryApiDeserializer();
-
-        okBuilder = new OkHttpClient.Builder();
-        addDefaultHeaderInterceptor();
```
jason-chan-lf, line 34: reminder to add back default header support later

src/main/java/com/laserfiche/repository/api/clients/impl/ApiClient.java
```diff
@@ -0,0 +1,15 @@
+package com.laserfiche.repository.api.clients.impl;
+
+import kong.unirest.UnirestInstance;
+
+public class ApiClient {
```
jason-chan-lf, line 5: class should be abstract?

src/main/java/com/laserfiche/repository/api/clients/impl/AttributesClientImpl.java
```diff
@@ -0,0 +1,95 @@
...
+
+    @Override
+    public CompletableFuture<ODataValueContextOfListOfAttribute> getTrusteeAttributeKeyValuePairs(String repoId,
+            Boolean everyone, String prefer, String select, String orderby, Integer top, Integer skip, Boolean count) {
+        Map<String, Object> queryParameters = new HashMap<>();
+        if (everyone != null) {
+            queryParameters.put("everyone", everyone);
+        }
+        if (select != null) {
+            queryParameters.put("select", select);
```
jason-chan-lf, line 59: odata query parameters missing dollar sign prefix `$select`. Does it still work?

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
```diff
@@ -0,0 +1,777 @@
...
+        if (culture != null) {
+            queryParameters.put("culture", culture);
+        }
+        return httpClient
+                .post(baseUrl + "/v1/Repositories/{repoId}/Entries/{parentEntryId}/{fileName}")
+                .routeParam("repoId", repoId)
+                .routeParam("parentEntryId", String.valueOf(parentEntryId))
+                .routeParam("fileName", fileName)
+                .queryString(queryParameters)
+                .contentType("application/json")
```
jason-chan-lf, line 125: does this import API work with content type as application/json? Because it's content type on swagger is multipart/form-data

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
```diff
@@ -0,0 +1,777 @@
...
+                    if (httpResponse.getStatus() == 429) {
+                        throw new RuntimeException("Rate limit is reached.");
+                    }
+                    return httpResponse.getBody();
+                });
+    }
+
+    @Override
+    public CompletableFuture<CreateEntryResult> importDocument(PostEntryWithEdocMetadataRequest requestBody,
+            String repoId, Integer parentEntryId, String fileName, Boolean autoRename, String culture) {
```
jason-chan-lf, line 111: is the electronicDocument file parameter missing?

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
```diff
@@ -0,0 +1,777 @@
...
+                    }
+                    if (httpResponse.getStatus() == 429) {
+                        throw new RuntimeException("Rate limit is reached.");
+                    }
+                    return httpResponse.getBody();
+                });
+    }
+
+    @Override
+    public CompletableFuture<CreateEntryResult> importDocument(PostEntryWithEdocMetadataRequest requestBody,
```
jason-chan-lf, line 110: Is the ordering of the parameters in this API different? Usually repoId is first but in this case the request body is first

src/main/java/com/laserfiche/repository/api/clients/impl/SimpleSearchesClientImpl.java
lf-jason-jia: Good catch. No, it will not handle it. The work is planned though (in TFS).

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
lf-jason-jia: Yes, it's different becasue the old code generator doesn't repsect the `x-position` property. With our own codegen, we got that part right.

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
lf-jason-jia: Yes, it's not working. It is planned work in another work item (TFS).

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
lf-jason-jia: Yes, it's not working. It is planned work in another work item (TFS).

src/main/java/com/laserfiche/repository/api/clients/impl/ApiClient.java
lf-jason-jia: No, ApiClient won't provide any method signature for anyone to implement. It will just contain concrete implementations.

src/main/java/com/laserfiche/repository/api/clients/impl/AttributesClientImpl.java
lf-jason-jia: I guess not. I will double check. I will update the code if it doesn't work.

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
lf-jason-jia: Thanks. @sxu1998  and I made a list. I will create stories/tasks based on it.

src/main/java/com/laserfiche/repository/api/clients/impl/model/PostEntryChildrenRequest.java
lf-jason-jia: It depends. So, if it's a public field and if access (getter) and mutation (setter) don't require extra operations, it's okay to do so via dot nodation. This is a compromise for the class to work with Boomi. Because if it contains private field, druing de/serialization, it will violate security rules.

src/main/java/com/laserfiche/repository/api/clients/impl/model/ODataValueContextOfIListOfContextHit.java
```diff
@@ -1,97 +1,15 @@
...
-   * (except the first line).
-   */
-  private String toIndentedString(java.lang.Object o) {
-    if (o == null) {
-      return "null";
-    }
-    return o.toString().replace("\n", "\n    ");
-  }
+    @JsonProperty("@odata.nextLink")
+    public String _atOdataNextLink = null;
```
jason-chan-lf, line 11: name of public property shouldnt have _ prefix

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
jason-chan-lf: this request body doesnt have x-position. there are some cases when x-position isnt there

jason-chan-lf (approved):
Approve to not block the PR. the comments will be resolved later

Issue #33, opened 2022-09-20
Added default header


steven-xu-lf:
set a default header to prevent the 429 issue from happening when running the integration tests 
also updated the accesskey function name reference
all the tests now should be able to run locally without any problems

PR #34, opened 2022-09-20
added test header for all integration tests

steven-xu-lf:
issue -> #33

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
+++ b/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
@@ -27,9 +27,10 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
     protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKey, String baseUrlDebug) {
         String baseUrl = baseUrlDebug != null ? baseUrlDebug : "https://api." + accessKey.domain + "/repository";
         httpClient = Unirest.spawnInstance();
+
         httpClient.config()
-               .setObjectMapper(new RepositoryClientObjectMapper())
-               .interceptor(new OAuthInterceptor(servicePrincipalKey, accessKey));
+                .setObjectMapper(new RepositoryClientObjectMapper())
+                .interceptor(new OAuthInterceptor(servicePrincipalKey, accessKey));
 
         attributesClient = new AttributesClientImpl(baseUrl, httpClient);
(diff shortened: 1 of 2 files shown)


src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -56,6 +56,8 @@ public static RepositoryApiClient CreateFromAccessKey(String servicePrincipalKey
     @Override
     public void setDefaultRequestHeaders(Map<String, String> defaultRequestHeaders) {
         defaultHeaders = defaultRequestHeaders;
+        String defaultHeadersKey = defaultHeaders.keySet().toArray()[0].toString();
+        httpClient.config().addDefaultHeader(defaultHeadersKey, defaultHeaders.get(defaultHeadersKey));
```
jason-chan-lf: If `defaultRequestHeaders` has multiple headers, only the first is added?

jason-chan-lf: if this function is called multiple times, will all headers be added as default or will it reset the previous default headers? For example if I call this function multiple times like this
`setDefaultRequestHeaders({ header1: value, header2: value })`
`setDefaultRequestHeaders({ header3: value })`
i guess i would expect only `header3` is sent as a default header since we dont have a way to reset the headers and this is how the other client libraries work.

steven-xu-lf: good catch, fixed

steven-xu-lf: after changing to function call to `setDefaultHeader`, you should only expect `header3` to be the default header

src/test/java/integration/BaseTest.java
```diff
@@ -47,7 +47,7 @@ public static void setUp() {
     }
 
     public static CompletableFuture<Entry> createEntry(RepositoryApiClient client, String entryName,
-            Integer parentEntryId, Boolean autoRename) {
+                                                       Integer parentEntryId, Boolean autoRename) {
```
samadpaydar3: I think the original form is better as it is the form what IDEs usually follow. No need to manually format the code.

jason-chan-lf (approved):
I approved, but better if @lf-jason-jia also takes a look

lf-jason-jia:
I did take a look and I think it's good. Sorry I should explicitly approve it earlier.

lf-jason-jia (approved):

Issue #36, opened 2022-09-21
Write more integration tests


steven-xu-lf:
kept 3 skipped tests due to the entry serialization error 
did not change the cancel operation test case due to the fact that another exception should be thrown later  
add the integration tests that are missing 
all tests should pass locally 
running tests should not cause side effects. If there is, e.g., having an entry created during test, it should be removed before the termination of the test run
added the serverSessionClient route to the class
refactor test case renaming convention
**Note:** Do not include foreach or nextlink integration tests since @samadpaydar3 is working on that 
**Note 2:** the current import document api in the client lib is missing the edoc parameter, so i did not include the import document test case but will include it later in another test case

PR #37, opened 2022-09-22
added and refactored integration tests

steven-xu-lf:
issue -> #36


src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -102,6 +104,9 @@ public SearchesClient getSearchesClient() {
         return searchesClient;
     }
 
+    @Override
+    public ServerSessionClient getServerSessionClient() { return serverSessionClient; }
```
lf-jason-jia: Server session is deprecated so we didn't add it to the repository client. I think all of its methods are essentially noops.

lf-jason-jia (dismissed):

lf-jason-jia (approved):

PR #38, opened 2022-09-23
Bug batch fix

lf-jason-jia:
In this PR, we fixed a bunch of bugs found in earlier version of the repository client.

1. We add a special case handling for the electronnic document upload.
2. For request body that doesn't have x-position specified, we moved it to the last.
3. For all OData parameters, we add a $ in front.
4. For routeParam and queryParam, we generate a helper function to filter out null ones and convert non-String one to their corresponding string value. We then include the output to the request.
5. For errors not specified in swagger.json, we also throw exceptions (such as 500 and 404).

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/clients/EntriesClient.java
+++ b/src/main/java/com/laserfiche/repository/api/clients/EntriesClient.java
@@ -3,6 +3,7 @@ package com.laserfiche.repository.api.clients;
 import com.laserfiche.repository.api.clients.impl.model.*;
 
 import java.io.File;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
@@ -100,7 +101,7 @@ public interface EntriesClient {
      * @return CompletableFuture<ODataValueOfIListOfWEntryLinkInfo> The return value
      */
     CompletableFuture<ODataValueOfIListOfWEntryLinkInfo> assignEntryLinks(String repoId, Integer entryId,
-            PutLinksRequest requestBody);
+            List<PutLinksRequest> requestBody);
 
     /**
      * - Assign a template to an entry.
--- a/src/main/java/com/laserfiche/repository/api/clients/EntriesClient.java
+++ b/src/main/java/com/laserfiche/repository/api/clients/EntriesClient.java
@@ -32,6 +32,8 @@ public interface EntriesClient {
             String prefer, Boolean formatValue, String culture, String select, String orderby, Integer top,
             Integer skip, Boolean count);
 
+    CompletableFuture<ODataValueContextOfIListOfFieldValue> getFieldValuesNextLink(String nextLink, int maxPageSize);
+
     /**
      * - Update the field values assigned to an entry.
      * - Provide the new field values to assign to the entry, and remove/reset all previously assigned field values.
@@ -52,7 +54,6 @@ public interface EntriesClient {
      * - Optionally sets metadata and electronic document component.
      * - Optional parameter: autoRename (default false). If an entry already exists with the given name, the entry will be automatically renamed. With this route, partial success is possible. The response returns multiple operation (entryCreate operation, setEdoc operation, setLinks operation, etc..) objects, which contain information about any errors that may have occurred during the creation. As long as the entryCreate operation succeeds, the entry will be created, even if all other operations fail.
      *
-     * @param requestBody   null
      * @param repoId        The requested repository ID.
      * @param parentEntryId The entry ID of the folder that the document will be created in.
      * @param fileName      The created document's file name.
@@ -60,10 +61,12 @@ public interface EntriesClient {
      *                      renamed if an entry already exists with the given name in the folder. The default value is false.
      * @param culture       An optional query parameter used to indicate the locale that should be used.
      *                      The value should be a standard language tag.
+     * @param file          The file that will be uploaded.
+     * @param requestBody   null
      * @return CompletableFuture<CreateEntryResult> The return value
      */
-    CompletableFuture<CreateEntryResult> importDocument(PostEntryWithEdocMetadataRequest requestBody, String repoId,
-            Integer parentEntryId, String fileName, Boolean autoRename, String culture);
+    CompletableFuture<CreateEntryResult> importDocument(String repoId, Integer parentEntryId, String fileName,
+            Boolean autoRename, String culture, File file, PostEntryWithEdocMetadataRequest requestBody);
 
     /**
      * - Returns the links assigned to an entry.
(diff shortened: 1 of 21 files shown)


src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
```diff
@@ -102,26 +99,24 @@ public CompletableFuture<ODataValueOfIListOfFieldValue> assignFieldValues(String
...
+        Map<String, Object> pathParameters = getNonNullParameters(new String[]{"repoId", "parentEntryId", "fileName"},
+                new Object[]{repoId, parentEntryId, fileName});
         return httpClient
                 .post(baseUrl + "/v1/Repositories/{repoId}/Entries/{parentEntryId}/{fileName}")
-                .routeParam("repoId", repoId)
-                .routeParam("parentEntryId", String.valueOf(parentEntryId))
-                .routeParam("fileName", fileName)
                 .queryString(queryParameters)
+                .routeParam(pathParameters)
                 .contentType("application/json")
```
jason-chan-lf: does import api work when content type isn't `multipart/form-data`?

src/main/java/com/laserfiche/repository/api/clients/EntriesClient.java
```diff
@@ -52,18 +54,19 @@ CompletableFuture<ODataValueOfIListOfFieldValue> assignFieldValues(String repoId
...
      * @param culture       An optional query parameter used to indicate the locale that should be used.
      *                      The value should be a standard language tag.
+     * @param file          The file that will be uploaded.
+     * @param requestBody   null
      * @return CompletableFuture<CreateEntryResult> The return value
      */
-    CompletableFuture<CreateEntryResult> importDocument(PostEntryWithEdocMetadataRequest requestBody, String repoId,
-            Integer parentEntryId, String fileName, Boolean autoRename, String culture);
+    CompletableFuture<CreateEntryResult> importDocument(String repoId, Integer parentEntryId, String fileName,
+            Boolean autoRename, String culture, File file, PostEntryWithEdocMetadataRequest requestBody);
```
jason-chan-lf, line 70: very minor, but should the `File file` parameter, be `File electronicDocument` to match swagger specs?

steven-xu-lf (commented):
since i do not think that i can comment on collapsed lines of unmodified code, should the parameter type in line 103 for the `assignEntryLinks` function in the `EntriesClient` interface (along with implementation) be changed to `List<PutLinksRequest>` ?

When i tried adding a test case for that, i would get a 400 if i just passed in a `PutLinksRequest` object but i was able to fix the test case after wrapping everything with the List wrapper in java 

for reference, I believe both the C# and js client repo libs have either `IEnumerable` interface wrapper of `PutLinkRequest` objects or an array of them

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
lf-jason-jia: No, it won't work. Our current tests haven't covered this yet (all existing tests pass). For this one specifically, let me update the code. The change didn't survive a earlier rebase (in the code generator project).

src/main/java/com/laserfiche/repository/api/clients/EntriesClient.java
lf-jason-jia: Yes.

lf-jason-jia:
> since i do not think that i can comment on collapsed lines of unmodified code, should the parameter type in line 103 for the assignEntryLinks function in the EntriesClient interface (along with implementation) be changed to List<PutLinksRequest> ?

> When i tried adding a test case for that, i would get a 400 if i just passed in a PutLinksRequest object but i was able to fix the test case after wrapping everything with the List wrapper in java

> for reference, I believe both the C# and js client repo libs have either IEnumerable interface wrapper of PutLinkRequest objects or an array of them


I think you are right. I will fix the code generator and update here.

lf-jason-jia:
@sxu1998 Hi Steven, thanks for pointing it out. I fixed the request body type.

steven-xu-lf:
> @sxu1998 Hi Steven, thanks for pointing it out. I fixed the request body type.

ok thanks!

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
lf-jason-jia: Fixed it. Thanks!

src/main/java/com/laserfiche/repository/api/clients/EntriesClient.java
lf-jason-jia: Fixed it. Thanks!

src/main/java/com/laserfiche/repository/api/clients/impl/TasksClientImpl.java
```diff
@@ -35,16 +37,20 @@ public CompletableFuture<OperationProgress> getOperationStatusAndProgress(String
...
                     }
+                    if (httpResponse.getStatus() >= 299) {
+                        throw new RuntimeException(httpResponse.getStatusText());
+                    }
                     return httpResponse.getBody();
                 });
     }
 
     @Override
     public CompletableFuture<Void> cancelOperation(String repoId, String operationToken) {
```
steven-xu-lf, line 48: just wondering for this specific API, should there an additional conditional statement added specifically for the `CancelOpertion_OperationEndedBeforeCancel` test case or an `ApiException` object be thrown instead? because right now it's just returning the generic runtime exception that prints out the message "Invalid or bad request", but for the cancel operation test case in the other 2 libraries, it is trying to assert this message below (which i don't believe that i can access in the java integration tests if just new RuntimeException("Invalid or bad request") is being thrown)

image

image

and after checking the implementation for both js and C#, it returns more than just the invalid or bad request message and it returns either an `ApiException<ProblemDetails>` or `ApiException` object with all those unique messages being set 

image


image

lf-jason-jia: Yeah, I think it's a good idea to include more details in the error reporting. We could do `ProblemDetails` but currently it's not on the priority list. But I can also include the exception in the report. I think that's something doesn't take much time to do while also bring enough details.

jason-chan-lf (dismissed):

src/main/java/com/laserfiche/repository/api/clients/impl/TasksClientImpl.java
steven-xu-lf: ok i just created a new low priority story for it -> https://tfs/DefaultCollection/Server%20Tools/_workitems/edit/397024 and i guess i will leave the cancel operation tests cases alone in my PR for now, thanks!

jason-chan-lf (approved):

PR #39, opened 2022-09-28
Model classes uses private fields with public accessors

lf-jason-jia:
1. All model classes now come with private fields and public accessors.
2. All model classes now has annotation that ignores null fields.
3. All tests have been updated and are passing (except fro nextLink ones, Samad will have a PR to fix it).
4. Removed forEach tests for now as the clients don't have forEach methods, again, Samad will have a PR soon to fix it.
5. Removed unused classes and imports.


src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
```diff
@@ -2,26 +2,175 @@
...
+    @Schema(description = "")
+    public String getOperationId() {
+        return operationId;
+    }
+
+    public void setOperationId(String operationId) {
+        this.operationId = operationId;
+    }
+
+    public APIServerException message(String message) {
```
samadpaydar3, line 45: It is not clear for me why do we need methods like **message** in addition to the corresponding setters/getters.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
```diff
@@ -2,26 +2,175 @@
...
+    public void setErrorSource(String errorSource) {
+        this.errorSource = errorSource;
+    }
+
+    @Override
+    public boolean equals(java.lang.Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
```
samadpaydar3, line 120: I think it's better to use instanceof (it also eliminates the need for null check)

if (!(o instanecOf APIServerException)) {
}

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
```diff
@@ -2,26 +2,175 @@
...
+    public int hashCode() {
+        return Objects.hash(operationId, message, errorCode, errorClass, statusCode, errorSource);
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("class APIServerException {\n");
+        sb
+                .append("    operationId: ")
```
samadpaydar3, line 141: minor but I think its better to have toString generate a less verbose representation. currently, the result includes multiple whitespaces, new lines...
I guess a comma separated list of attribute name/values is better.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
```diff
@@ -2,26 +2,175 @@
...
+                .append(toIndentedString(operationId))
+                .append("\n");
+        sb
```
samadpaydar3, line 144: Isn't it better to update the code generator to not break the statement to multiple lines? I think the generated method body is longer than needed.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
```diff
@@ -2,26 +2,175 @@
...
+    public String getErrorSource() {
+        return errorSource;
+    }
+
+    public void setErrorSource(String errorSource) {
+        this.errorSource = errorSource;
+    }
+
+    @Override
+    public boolean equals(java.lang.Object o) {
```
samadpaydar3, line 116: maybe it's better to use only "Object" instead of the fully qualified name "java.lang.Object".

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
```diff
@@ -2,26 +2,175 @@
...
+                .append("    message: ")
+                .append(toIndentedString(message))
+                .append("\n");
+        sb
+                .append("    errorCode: ")
+                .append(toIndentedString(errorCode))
+                .append("\n");
+        sb
+                .append("    errorClass: ")
+                .append(toIndentedString(errorClass))
```
samadpaydar3, line 154: Maybe it's better to update the code generator to generate a more compact code like
sb.append(String.format("    errorClass: %s%n"), toIndentedString(errorClass));

or 

sb.append(String.format("    errorClass: %s%n"), String.valueOf(errorClass));

This also removes the toIndentedString which currently is duplicated in all the classes.

src/test/java/integration/BaseTest.java
```diff
@@ -58,7 +58,7 @@ public static CompletableFuture<Entry> createEntry(RepositoryApiClient client, S
 
     public static CompletableFuture<Boolean> allFalse(List<TemplateFieldInfo> arr) {
         for (TemplateFieldInfo templateFieldInfo : arr) {
-            if (templateFieldInfo.isRequired) {
+            if (templateFieldInfo.isIsRequired()) {
```
samadpaydar3, line 61: the generated method name is not appropriate. maybe we can check that if a boolean property starts with is, then don't add is to the getter method name.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
lf-jason-jia: Hi Samad, it's the generated code from Swagger Codegen. But I do agree, it's weird. Do you think we should have a custom transformation in the preprocessing project to remove such methods?

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
lf-jason-jia: Hi Samad, all these code under /model are auto-generated by Swagger Codegen.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
lf-jason-jia: Again, the code is from Swagger Code gen. But yeah, I agree that it could be done better. I think we can probably have a story that eliminates dependency on Swagger Code in the future.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
lf-jason-jia: The code is actually not generated by our code generator but Swagger Codogen. So I guess no?

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
lf-jason-jia: Hi Samad, this code is from Swagger codegen and they are template-based code generators unlike us. I think they don't have a safe way of resolving import statements so they did the lazy way: fully qualified name.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
lf-jason-jia: I agree. I think it could be a future story (the current one works).

src/test/java/integration/BaseTest.java
lf-jason-jia: Indeed, I feel this too. I think Swagger codegen's way of doing this is: if it's a boolean, use "Is" prefix.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
samadpaydar3: Yes. I think it is needed, as having this kind of method is not common in java.
Maybe we can create a new user story for removing such methods.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
samadpaydar3: Oh, sorry, I didn't notice that.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
samadpaydar3: Oh, sorry. 
Yeah, I think it's totally ok for now.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
samadpaydar3: yeah, I didn't noticed it's generated by swagger codegen. yeah, It's totally ok.

src/main/java/com/laserfiche/repository/api/clients/impl/model/APIServerException.java
samadpaydar3: I see. yeah. Thanks for the feedback.

samadpaydar3 (approved):

src/test/java/integration/BaseTest.java
samadpaydar3: yeah. I see.

PR #40, opened 2022-09-29
Story/392570 nextlink foreach codegen

samadpaydar3:
Fix issues in integration tests.
Update auto-generated code (nextLink and forEach methods)


.github/workflows/main.yml
```diff
@@ -17,63 +17,63 @@ jobs:
...
-        REPOSITORY_ID:  ${{ secrets.DEV_CA_PUBLIC_USE_REPOSITORY_ID_1 }}
-        TEST_HEADER: ${{ secrets.TEST_HEADER }}
-      run: mvn test
-
-    - name: Publish Test Results
-      uses: EnricoMi/publish-unit-test-result-action@v1
-      if: always()
-      with:
-        files: "target/surefire-reports/*.xml"
+      - uses: actions/checkout@v3
```
lf-jason-jia: This is just an indentation change. Do we want to change the indentation?

lf-jason-jia (changes requested):
See comments

DEVELOPER.md
```diff
@@ -6,33 +6,54 @@ flowchart TD
...
-|OAuth Service|Laserfiche OAuth service|See [lf-api-client-core-java](https://github.com/Laserfiche/lf-api-client-core-java)|N/A|
-|Custom Deserialization|Ideally we would rely on the auto generated code but it fails to handle anyOf, oneOf, allOf so we have our own code for it.|Anything under com.laserfiche.repository.api.serialization|No interface|
+|RepositoryApiClient|This is the public interface the developer uses to get data from and set value to the Laserfiche
+Repository|RepositoryApiClient|RepositoryApiClientImpl|
+|OAuthInterceptor|It uses [lf-api-client-core-java](https://github.com/Laserfiche/lf-api-client-core-java) to get access
+token from the OAuth service, and converts the token into a Bearer header for each outgoing HTTP
+request.|okhttp3.interceptor|OAuthInterceptor|
+|[Retrofit2](https://square.github.io/retrofit/)/[OkHttp3](https://square.github.io/okhttp/)|We use Swagger Codegen to
```
lf-jason-jia: Is this outdated? It mentions Retrofit which we no longer use.

README.md
```diff
@@ -1,18 +1,24 @@
 # Laserfiche Repository API Client
 
-Use the Laserfiche Repository API to access data in a Laserfiche repository. Import or export files, modify the repository folder structure, read and modify templates and field values, and more.
+Use the Laserfiche Repository API to access data in a Laserfiche repository. Import or export files, modify the
+repository folder structure, read and modify templates and field values, and more.
```
lf-jason-jia: Do we want to break this paragraph into two?

pom.xml
```diff
@@ -1,313 +1,313 @@
...
+    </prerequisites>
 
-  <licenses>
-    <license>
-      <name>MIT</name>
-      <url>https://opensource.org/licenses/MIT</url>
-      <distribution>repo</distribution>
-    </license>
-  </licenses>
+    <licenses>
```
lf-jason-jia: Again, do we want to change the indentation?

src/main/java/com/laserfiche/repository/api/clients/AttributesClient.java
```diff
@@ -1,9 +1,16 @@
 package com.laserfiche.repository.api.clients;
 
-import com.laserfiche.repository.api.clients.impl.model.Attribute;
-import com.laserfiche.repository.api.clients.impl.model.ODataValueContextOfListOfAttribute;
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
```
lf-jason-jia: We probably don't need these imports for interfaces. I forgot to mention that you could run "optimize imports" in batch to remove unused imports.

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
```diff
@@ -353,10 +410,15 @@ public CompletableFuture<File> exportDocumentWithAuditReason(String repoId, Inte
     public CompletableFuture<File> exportDocument(String repoId, Integer entryId, String range) {
         Map<String, Object> pathParameters = getNonNullParameters(new String[]{"repoId", "entryId"},
                 new Object[]{repoId, entryId});
+        Map<String, Object> headerParameters = getNonNullParameters(new String[]{"range"}, new Object[]{range});
+        Map<String, String> headerParametersWithStringTypeValue = headerParameters
+                .entrySet()
+                .stream()
+                .collect(Collectors.toMap(Map.Entry::getKey, e -> (String) e.getValue()));
```
lf-jason-jia, line 415: This seems really nice. I know it's not the job of this PR, but maybe we could have story that convert the other two use cases (in header and in path) to use this instead.

src/main/java/com/laserfiche/repository/api/clients/impl/model/FuzzyType.java
```diff
@@ -2,6 +2,8 @@
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
+import com.fasterxml.jackson.annotation.JsonInclude;
```
lf-jason-jia: Again, we could batch run "optimize imports" to remove unused imports. Ideally, we could have the ability to resolve dependencies during code generation, but I expect it will be tough (the code gen probably will need to have knowledge of the entire project to do the resolution correctly, I am not sure).

src/main/java/com/laserfiche/repository/api/clients/impl/SearchesClientImpl.java
```diff
@@ -123,11 +134,30 @@ private CompletableFuture<ODataValueContextOfIListOfContextHit> doGetSearchConte
...
+    public CompletableFuture<Void> getSearchContextHitsForEach(
+            Function<CompletableFuture<ODataValueContextOfIListOfContextHit>, CompletableFuture<Boolean>> callback,
+            Integer maxPageSize, String repoId, String searchToken, Integer rowNumber, String prefer, String select,
+            String orderby, Integer top, Integer skip, Boolean count) throws InterruptedException, ExecutionException {
+        prefer = mergeMaxSizeIntoPrefer(maxPageSize, prefer);
+        CompletableFuture<ODataValueContextOfIListOfContextHit> response = getSearchContextHits(repoId, searchToken,
+                rowNumber, prefer, select, orderby, top, skip, count);
+        while (response != null && callback
+                .apply(response)
+                .get()) {
```
lf-jason-jia: If we use `join()`, would it save us those: `throws InterruptedException, ExecutionException`?

lf-jason-jia: Reference: https://stackoverflow.com/questions/45490316/completablefuturet-class-join-vs-get

.github/workflows/main.yml
samadpaydar3: Sorry. This was unintended. I wanted to reformat java code, and by mistake have reformatted all the project files.
I fixed it.

DEVELOPER.md
samadpaydar3: The original file is replaced.

README.md
samadpaydar3: No. this was done by mistake. I have replaced the original file.

pom.xml
samadpaydar3: I replaced the original file.

src/main/java/com/laserfiche/repository/api/clients/AttributesClient.java
samadpaydar3: I removed unnecessary imports form clients (interfaces and their implementations).

src/main/java/com/laserfiche/repository/api/clients/impl/SearchesClientImpl.java
samadpaydar3: both methods potentially throw exceptions, but join's exceptions are unchecked, while get's exceptions are checked. http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/java/util/concurrent/CompletableFuture.java
So, using join, we don't need the throws declaration. However I am not sure which one is better.
Currently, when a client calls this method in their code, as the method throws checked exceptions, the client is alerted by the compiler and need to make their code safe against these exceptions, if they are thrown. So, while this might be a burden for them, it helps them have more robust code. 
If we use join, then it might happen that a client who uses this method is not aware that what kinds of exceptions might occur, and do not use any exception handling, reducing their code's robustness. 
I'm ok with both but are personally prefer throwing checked exceptions. It's good to mob with Paolo about this.

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
samadpaydar3: I created this user story https://tfs/defaultcollection/Server%20Tools/_workitems/edit/397778 for this.

src/main/java/com/laserfiche/repository/api/clients/impl/model/FuzzyType.java
samadpaydar3: Thanks for mentioning the "optimize imports" in intellij. I wasn't familiar with that. I used it to remove unnecessary imports.

src/main/java/com/laserfiche/repository/api/clients/impl/SearchesClientImpl.java
samadpaydar3: I noticed that other methods (in the autogenerated interfaces) do not throw checked exceptions. So, I will change the code to use join, so that forEach methods are also like others.

lf-jason-jia (approved):

PR #45, opened 2022-10-06
Add Subtype Hint Annotations

lf-jason-jia:
When an API that returns an Entry object, it could mean any of its subtypes: Shortcut and Folder. But the client deserializer has no way of knowing this so we add the type hint here.

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/clients/impl/model/Entry.java
+++ b/src/main/java/com/laserfiche/repository/api/clients/impl/model/Entry.java
@@ -2,6 +2,8 @@ package com.laserfiche.repository.api.clients.impl.model;
 
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import io.swagger.v3.oas.annotations.media.Schema;
 import org.threeten.bp.OffsetDateTime;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/laserfiche/repository/api/clients/impl/model/Entry.java
```diff
@@ -10,6 +12,8 @@
 import java.util.Objects;
 
 @javax.annotation.Generated(value = "io.swagger.codegen.v3.generators.java.JavaClientCodegen", date = "2022-08-19T17:07:43.799-04:00[America/New_York]")
+@JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION)
+@JsonSubTypes({@JsonSubTypes.Type(Folder.class), @JsonSubTypes.Type(Shortcut.class)})
```
jason-chan-lf: Should `Document.class` also be added? Reminder that in the future `RecordSeries.class` too

lf-jason-jia: Oh, OK. I didn't know about it. Let me add both.

lf-jason-jia: I will also upodate the code generator PR.

lf-jason-jia: I just checked the code, yes, we will add `RecordSeries.class` when we have it. For now, we will just add `Document.class`.

samadpaydar3: I know that the tests pass, but just to improve the code, isn't it better to use fully qualified class names. I am not sure how Folder.class is resolved.

jason-chan-lf (approved):

src/main/java/com/laserfiche/repository/api/clients/impl/model/Entry.java
lf-jason-jia: Hi Samad, that's a good catch. But I think I know how it's resolved: in this case, it doesn't need an import statement or fully qualified class name because the Entry class and its subtypes are in the same package.

samadpaydar3 (approved):

PR #46, opened 2022-10-11
Story/397206 file upload parameter

samadpaydar3:
Import api is updated to use InputStream instead of File, as the parameter
integration tests for the import api are updated
new integration tests are added for importing from url and a string object.

### DIFF
--- a/src/test/java/integration/ImportDocumentApiTest.java
+++ b/src/test/java/integration/ImportDocumentApiTest.java
@@ -1,9 +1,20 @@
 package integration;
 
 import com.laserfiche.repository.api.clients.EntriesClient;
-import com.laserfiche.repository.api.clients.impl.model.DeleteEntryWithAuditReason;
+import com.laserfiche.repository.api.clients.impl.model.*;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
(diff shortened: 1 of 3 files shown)


src/test/java/integration/ImportDocumentApiTest.java
```diff
@@ -1,9 +1,20 @@
...
 import com.laserfiche.repository.api.clients.EntriesClient;
-import com.laserfiche.repository.api.clients.impl.model.DeleteEntryWithAuditReason;
+import com.laserfiche.repository.api.clients.impl.model.*;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.io.*;
+import java.net.ServerSocket;
```
lf-jason-jia: Nitpick: it seems we don't need this import.

lf-jason-jia (dismissed):

src/test/java/integration/ImportDocumentApiTest.java
samadpaydar3: Fixed.

lf-jason-jia (approved):

PR #48, opened 2022-10-13
Subtype casting

lf-jason-jia:
1. Fix the deserialization annotation so that the incoming JSON will be converted to the subtypes correctly.
2. Update the test to make sure the conversion is correct.

### DIFF
--- a/src/test/java/integration/EntriesApiTest.java
+++ b/src/test/java/integration/EntriesApiTest.java
@@ -58,8 +58,6 @@ class EntriesApiTest extends BaseTest {
                     break;
             }
         }
-
-        assertTrue(true);
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


src/test/java/integration/EntriesApiTest.java
```diff
@@ -39,10 +39,27 @@ void getEntry_ReturnRootFolder() {
...
+                case SHORTCUT:
+                    assertTrue(entry instanceof Shortcut);
+                    break;
+                case DOCUMENT:
+                    assertTrue(entry instanceof Document);
+                    break;
+            }
+        }
+
+        assertTrue(true);
```
samadpaydar3: Do we need this?

samadpaydar3 (dismissed):

src/test/java/integration/EntriesApiTest.java
lf-jason-jia: Oh, we don't. Let me update the code.

samadpaydar3 (approved):

Issue #52, opened 2022-10-18
Update client core dependency version


steven-xu-lf:
updated client core dependency version in the pom.xml file in after publishing preview
updated repo client
after changing fields to private and adding accessors
**Note:** majority of the integration tests pass locally for me, the ones that do not pass i believe are all unrelated to this update as they were here before this change (will probably need to log a bug for this)

PR #53, opened 2022-10-18
updated client core dependency

steven-xu-lf:
issue -> #52

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -263,7 +263,7 @@
     <dependency>
       <groupId>com.laserfiche</groupId>
       <artifactId>lf-api-client-core</artifactId>
-      <version>1.0.0</version>
+      <version>1.0.0-preview-3274168162-SNAPSHOT</version>
     </dependency>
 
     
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -263,7 +263,7 @@
     <dependency>
       <groupId>com.laserfiche</groupId>
       <artifactId>lf-api-client-core</artifactId>
-      <version>1.0.0</version>
```
lf-jason-jia: Is this published to sonar type snapshots?

steven-xu-lf: yes, after running the publish preview step in the client core pipeline -> https://s01.oss.sonatype.org/content/repositories/snapshots/com/laserfiche/lf-api-client-core/1.0.0-preview-3274168162-SNAPSHOT/

lf-jason-jia: ok

lf-jason-jia: Can we keep it 1.0.0 in the 1.x branch? The reason is, we have decided to not to work on the pipeline and focus on providing everything for the Integration team. Keeping the version in 1.x as 1.0.0 (so not using the snapshot) would be easier since we haven't resolved the firewall issue (again, we decided to postpone it for support the Integration team).

lf-jason-jia (changes requested):
See comments

lf-jason-jia (approved):

PR #62, opened 2022-11-01
add selfhosted implementation

steven-xu-lf:
add a function that initializes the repository client from user name and password
Add a self hosted interceptor object that implements the interceptor interface
Make sure repo client uses the APIException, and the OffsetDateTimeDeserializer class from the client core repo
configure baseTest to be able to run all the same integration tests on selfhosted
test locally 

Reference -> https://github.com/Laserfiche/lf-repository-api-client-dotnet/blob/1.x/src/RepositoryApiClient.cs

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
+++ b/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
@@ -62,9 +62,9 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
         return createFromAccessKey(servicePrincipalKey, accessKey, null);
     }
 
-    public static RepositoryApiClient createFromUsernamePassword(String repoId, String username, String password,
+    public static RepositoryApiClient createFromUsernamePassword(String repositoryId, String username, String password,
             String baseUrl) {
-        Interceptor interceptor = new SelfHostedInterceptor(repoId, username, password, baseUrl, null);
+        Interceptor interceptor = new SelfHostedInterceptor(repositoryId, username, password, baseUrl, null);
         return new RepositoryApiClientImpl(interceptor, baseUrl);
     }
 
--- a/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
+++ b/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
@@ -12,7 +12,7 @@ import java.util.Map;
 
 public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseable {
     private Map<String, String> defaultHeaders;
-    private static UnirestInstance httpClient;
+    private static UnirestInstance httpClient = Unirest.spawnInstance();
     private final AttributesClient attributesClient;
     private final AuditReasonsClient auditReasonsClient;
     private final EntriesClient entriesClient;
@@ -44,7 +44,6 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
     public static RepositoryApiClient CreateFromHttpRequestHandler(String servicePrincipalKey, AccessKey accessKey,
             String baseUrl) {
         String baseUrlDebug = baseUrl != null ? baseUrl : "https://api." + accessKey.getDomain() + "/repository";
-        httpClient = Unirest.spawnInstance();
         httpClient
                 .config()
                 .setObjectMapper(new RepositoryClientObjectMapper())
@@ -55,7 +54,6 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
     public static RepositoryApiClient CreateFromHttpRequestHandler(String repositoryId, String username,
             String password,
             TokenClient client, String baseUrl) {
-        httpClient = Unirest.spawnInstance();
         httpClient
                 .config()
                 .setObjectMapper(new RepositoryClientObjectMapper())
--- a/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
+++ b/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
@@ -1,10 +1,10 @@
 package com.laserfiche.repository.api;
 
-import com.laserfiche.api.client.apiserver.TokenClient;
 import com.laserfiche.api.client.model.AccessKey;
 import com.laserfiche.repository.api.clients.*;
 import com.laserfiche.repository.api.clients.impl.*;
 import com.laserfiche.repository.api.clients.impl.deserialization.RepositoryClientObjectMapper;
+import kong.unirest.Interceptor;
 import kong.unirest.Unirest;
 import kong.unirest.UnirestInstance;
 
@@ -26,26 +26,15 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
     private final TasksClient tasksClient;
     private final TemplateDefinitionsClient templateDefinitionsClient;
 
-    protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKey, String repositoryId,
-            String username,
-            String password, TokenClient client, String baseUrl) {
-        httpClient = Unirest.spawnInstance();
-        if (baseUrl == null) {
-            baseUrl = "https://api." + accessKey.getDomain() + "/repository";
-        } else if (baseUrl.endsWith("/")) {
+    protected RepositoryApiClientImpl(Interceptor interceptor, String baseUrl) {
+        if (baseUrl.endsWith("/")) {
             baseUrl = baseUrl.substring(0, baseUrl.length() - 1);
         }
-        if (servicePrincipalKey == null && accessKey == null) {
-            httpClient
-                    .config()
-                    .setObjectMapper(new RepositoryClientObjectMapper())
-                    .interceptor(new SelfHostedInterceptor(repositoryId, username, password, baseUrl, client));
-        } else {
-            httpClient
-                    .config()
-                    .setObjectMapper(new RepositoryClientObjectMapper())
-                    .interceptor(new OAuthInterceptor(servicePrincipalKey, accessKey));
-        }
+        httpClient = Unirest.spawnInstance();
+        httpClient
+                .config()
+                .setObjectMapper(new RepositoryClientObjectMapper())
+                .interceptor(interceptor);
         attributesClient = new AttributesClientImpl(baseUrl, httpClient);
         auditReasonsClient = new AuditReasonsClientImpl(baseUrl, httpClient);
         entriesClient = new EntriesClientImpl(baseUrl, httpClient);
@@ -62,7 +51,11 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
 
     public static RepositoryApiClient createFromAccessKey(String servicePrincipalKey, AccessKey accessKey,
             String baseUrlDebug) {
-        return new RepositoryApiClientImpl(servicePrincipalKey, accessKey, null, null, null, null, baseUrlDebug);
+        if (baseUrlDebug == null) {
+            baseUrlDebug = "https://api." + accessKey.getDomain() + "/repository";
+        }
+        Interceptor interceptor = new OAuthInterceptor(servicePrincipalKey, accessKey);
+        return new RepositoryApiClientImpl(interceptor, baseUrlDebug);
     }
 
     public static RepositoryApiClient createFromAccessKey(String servicePrincipalKey, AccessKey accessKey) {
@@ -71,7 +64,8 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
 
     public static RepositoryApiClient createFromUsernamePassword(String repoId, String username, String password,
             String baseUrl) {
-        return new RepositoryApiClientImpl(null, null, repoId, username, password, null, baseUrl);
+        Interceptor interceptor = new SelfHostedInterceptor(repoId, username, password, baseUrl, null);
+        return new RepositoryApiClientImpl(interceptor, baseUrl);
     }
 
     @Override
--- a/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
+++ b/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
@@ -28,13 +28,14 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
 
     protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKey, String repositoryId,
             String username,
-            String password, TokenClient client, String baseUrl) {
+            String password, TokenClient client, String baseUrlDebug) {
         httpClient = Unirest.spawnInstance();
+        String baseUrl = baseUrlDebug != null ? baseUrlDebug : "https://api." + accessKey.getDomain() + "/repository";
         if (servicePrincipalKey == null && accessKey == null) {
             httpClient
                     .config()
                     .setObjectMapper(new RepositoryClientObjectMapper())
-                    .interceptor(new SelfHostedInterceptor(repositoryId, username, password, baseUrl, client));
+                    .interceptor(new SelfHostedInterceptor(repositoryId, username, password, baseUrlDebug, client));
         } else {
             httpClient
                     .config()
--- a/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
+++ b/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
@@ -41,7 +41,7 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
         templateDefinitionsClient = new TemplateDefinitionsClientImpl(baseUrl, httpClient);
     }
 
-    public static RepositoryApiClient CreateFromHttpRequestHandler(String servicePrincipalKey, AccessKey accessKey,
+    public static RepositoryApiClient createFromHttpRequestHandler(String servicePrincipalKey, AccessKey accessKey,
             String baseUrl) {
         String baseUrlDebug = baseUrl != null ? baseUrl : "https://api." + accessKey.getDomain() + "/repository";
         httpClient
@@ -51,7 +51,7 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
         return new RepositoryApiClientImpl(httpClient, baseUrlDebug);
     }
 
-    public static RepositoryApiClient CreateFromHttpRequestHandler(String repositoryId, String username,
+    public static RepositoryApiClient createFromHttpRequestHandler(String repositoryId, String username,
             String password,
             TokenClient client, String baseUrl) {
         httpClient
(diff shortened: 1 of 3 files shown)


src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -48,13 +41,34 @@ protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKe
         templateDefinitionsClient = new TemplateDefinitionsClientImpl(baseUrl, httpClient);
     }
 
-    public static RepositoryApiClient CreateFromAccessKey(String servicePrincipalKey, AccessKey accessKey,
-            String baseUrlDebug) {
-        return new RepositoryApiClientImpl(servicePrincipalKey, accessKey, baseUrlDebug);
+    public static RepositoryApiClient CreateFromHttpRequestHandler(String servicePrincipalKey, AccessKey accessKey,
```
lf-jason-jia: As we just discussed, method name starts with lower case letter.

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -11,7 +12,7 @@
 
 public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseable {
     private Map<String, String> defaultHeaders;
-    private UnirestInstance httpClient;
+    private static UnirestInstance httpClient = Unirest.spawnInstance();
```
lf-jason-jia: Actually, while we leave out the Unirest single issue out of this PR, this line does change the behavior:
- Old behavior: all instances each has an instance of Unirest.
- New behavior: all instances of `RepositoryApiClientImpl` shares one instance of Unirest.

Not that the new behavior is bad, but it's wasteful. Each Unirest instance would have a thread pool, which is unnecessary and kind of negate the benefits of async pattern (because we could have just get a thread for each API call if resource isn't a concern)

lf-jason-jia: So I think we could keep the old behavior and also keep the part where we initialize the client in the constructor.

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -48,13 +41,34 @@ protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKe
...
+            String baseUrl) {
+        String baseUrlDebug = baseUrl != null ? baseUrl : "https://api." + accessKey.getDomain() + "/repository";
+        httpClient
+                .config()
+                .setObjectMapper(new RepositoryClientObjectMapper())
+                .interceptor(new OAuthInterceptor(servicePrincipalKey, accessKey));
+        return new RepositoryApiClientImpl(httpClient, baseUrlDebug);
+    }
+
+    public static RepositoryApiClient CreateFromHttpRequestHandler(String repositoryId, String username,
```
lf-jason-jia: Same, naming convention

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -48,13 +41,34 @@ protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKe
...
+            String password,
+            TokenClient client, String baseUrl) {
+        httpClient
+                .config()
+                .setObjectMapper(new RepositoryClientObjectMapper())
+                .interceptor(new SelfHostedInterceptor(repositoryId, username, password, baseUrl, client));
+        return new RepositoryApiClientImpl(httpClient, baseUrl);
     }
 
     public static RepositoryApiClient CreateFromAccessKey(String servicePrincipalKey, AccessKey accessKey) {
```
lf-jason-jia: Naming convention

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -48,13 +41,34 @@ protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKe
...
-        return CreateFromAccessKey(servicePrincipalKey, accessKey, null);
+        return CreateFromHttpRequestHandler(servicePrincipalKey, accessKey, null);
+    }
+
+    public static RepositoryApiClient CreateFromUsernamePassword(String repoId, String username, String password,
```
lf-jason-jia: Naming convention

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -48,13 +41,34 @@ protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKe
...
+                .setObjectMapper(new RepositoryClientObjectMapper())
+                .interceptor(new OAuthInterceptor(servicePrincipalKey, accessKey));
+        return new RepositoryApiClientImpl(httpClient, baseUrlDebug);
+    }
+
+    public static RepositoryApiClient CreateFromHttpRequestHandler(String repositoryId, String username,
+            String password,
+            TokenClient client, String baseUrl) {
+        httpClient
+                .config()
```
lf-jason-jia: Can we move this back to the constructor, it causes code duplication.

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -48,13 +41,34 @@ protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKe
...
     }
 
     public static RepositoryApiClient CreateFromAccessKey(String servicePrincipalKey, AccessKey accessKey) {
-        return CreateFromAccessKey(servicePrincipalKey, accessKey, null);
+        return CreateFromHttpRequestHandler(servicePrincipalKey, accessKey, null);
+    }
+
+    public static RepositoryApiClient CreateFromUsernamePassword(String repoId, String username, String password,
+            String baseUrl) {
+        String baseUrlWithSlash = baseUrl.endsWith("/") ? baseUrl.substring(0, baseUrl.lastIndexOf("/")) : baseUrl;
```
lf-jason-jia: Why use `lastIndexOf` if we already know it's at the end. BTW, `lastIndexOf` is O(n) where n is the number of characters in the string.

lf-jason-jia (dismissed):
See comments

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
sxu1998: yea so this was put to account for the edge case of having a backslash at the end of the self hosted base URL so for example if i do not have this:
`https://v-dev-w10x64-73.laserfiche.com/LFRepositoryAPI/` will throw an error even tho it's not supposed to

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
steven-xu-lf: ok so in this case, i believe i would have to get rid of the both `createFromHttpRequestHandler` functions, since the whole purpose of both those functions is to create an instance and initialize `httpClient` and add an extra conditional case for selfhosted in the constructor

src/test/java/integration/BaseTest.java
```diff
@@ -14,36 +14,76 @@
...
                     .filename(".env")
                     .load();
-            accessKeyBase64 = dotenv.get("ACCESS_KEY");
-            spKey = dotenv.get("SERVICE_PRINCIPAL_KEY");
-            repoId = dotenv.get("REPOSITORY_ID");
+            authorizationType = dotenv.get("AUTHORIZATION_TYPE");
             testHeaderValue = dotenv.get("TEST_HEADER");
+            if (authorizationType.equalsIgnoreCase("CloudAccessKey")) {
+                if (nullOrEmpty(spKey) && nullOrEmpty(accessKeyBase64) && nullOrEmpty(repoId) && nullOrEmpty(
+                        testHeaderValue)) {
```
jason-chan-lf: `testHeaderValue` doesn't need to be checked if null or empty here since it isn't being assigned?

src/test/java/integration/BaseTest.java
```diff
@@ -14,36 +14,76 @@
...
+                    accessKeyBase64 = dotenv.get("ACCESS_KEY");
+                    spKey = dotenv.get("SERVICE_PRINCIPAL_KEY");
+                    repoId = dotenv.get("REPOSITORY_ID");
```
jason-chan-lf: `repoId` is used by both authorization types, so could be moved out of these `if` statements

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -25,15 +26,21 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
     private final TasksClient tasksClient;
     private final TemplateDefinitionsClient templateDefinitionsClient;
 
-    protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKey, String baseUrlDebug) {
-        String baseUrl = baseUrlDebug != null ? baseUrlDebug : "https://api." + accessKey.getDomain() + "/repository";
```
jason-chan-lf, line 29: does cloud api still works with this removed?

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -48,13 +55,19 @@ protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKe
...
+
+    public static RepositoryApiClient createFromAccessKey(String servicePrincipalKey, AccessKey accessKey) {
+        return createFromAccessKey(servicePrincipalKey, accessKey, null);
     }
 
-    public static RepositoryApiClient CreateFromAccessKey(String servicePrincipalKey, AccessKey accessKey) {
-        return CreateFromAccessKey(servicePrincipalKey, accessKey, null);
+    public static RepositoryApiClient createFromUsernamePassword(String repoId, String username, String password,
+            String baseUrl) {
+        String baseUrlWithSlash = baseUrl.endsWith("/") ? baseUrl.substring(0, baseUrl.length() - 1) : baseUrl;
```
jason-chan-lf: could move this line that trims the baseUrl to `RepositoryApiClientImpl` since it also has the baseUrl parameter and if `createFromAccessKey(String servicePrincipalKey, AccessKey accessKey, String baseUrlDebug)` is called, the trim functionality would be reused

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
steven-xu-lf: oh shoot forgot about that, thanks for reminder

src/test/java/integration/BaseTest.java
steven-xu-lf: ok done and also thrown an exception if `repoId` is not found

src/test/java/integration/BaseTest.java
```diff
@@ -14,7 +14,12 @@
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-//use enumerators
+
+enum authorizationTypes {
```
jason-chan-lf: is Java conventions uppercase for enum type name? Also, should be singular case? `AuthorizationType.CLOUDACCESSKEY` vs `AuthorizationTypes.CLOUDACCESSKEY`

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -11,7 +12,7 @@
 
 public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseable {
     private Map<String, String> defaultHeaders;
-    private UnirestInstance httpClient;
+    private static UnirestInstance httpClient;
```
jason-chan-lf: since this is static, does that mean if I create two `RepositoryApiClientImpl`, the httpClient is shared? And if I call `setDefaultRequestHeaders` on one `RepositoryApiClientImpl`, it will also affect the second `RepositoryApiClientImpl`?

src/test/java/integration/BaseTest.java
steven-xu-lf: yes because they are `static final`, and changed it to singular case

jason-chan-lf: what about the enum type name `authorizationType` vs `AuthorizationType`? I see most places are uppercase first letter convention

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
steven-xu-lf: ok done

src/test/java/integration/BaseTest.java
steven-xu-lf: ok done

lf-jason-jia: Please add underscore for better readability.

steven-xu-lf: ok done

src/test/java/integration/BaseTest.java
```diff
@@ -15,35 +15,81 @@
...
+        repositoryApiClient = createClient();
+    }
+
+    public static RepositoryApiClient createClient() {
+        if (repositoryApiClient == null) {
+            if (authorizationType.equalsIgnoreCase("Cloud_Access_Key")) {
+                if (nullOrEmpty(spKey) || accessKey == null)
+                    return null;
+                repositoryApiClient = RepositoryApiClientImpl.createFromAccessKey(spKey, accessKey);
+            } else if (authorizationType.equalsIgnoreCase("API_Server_Username_Password")) {
```
jason-chan-lf: should use the enum instead of hardcode string `"API_Server_Username_Password"`. Same for above `"Cloud_Access_Key"` on line 80

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -48,13 +60,19 @@ protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKe
...
+
+    public static RepositoryApiClient createFromAccessKey(String servicePrincipalKey, AccessKey accessKey) {
+        return createFromAccessKey(servicePrincipalKey, accessKey, null);
     }
 
-    public static RepositoryApiClient CreateFromAccessKey(String servicePrincipalKey, AccessKey accessKey) {
-        return CreateFromAccessKey(servicePrincipalKey, accessKey, null);
+    public static RepositoryApiClient createFromUsernamePassword(String repoId, String username, String password,
+            String baseUrl) {
+        String baseUrlWithSlash = baseUrl.endsWith("/") ? baseUrl.substring(0, baseUrl.length() - 1) : baseUrl;
```
jason-chan-lf: can remove this line that removes trailing slash from url now that it is being done in the constructor?

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -25,15 +26,26 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
...
-
-        httpClient
-                .config()
-                .setObjectMapper(new RepositoryClientObjectMapper())
-                .interceptor(new OAuthInterceptor(servicePrincipalKey, accessKey));
-
+        if (baseUrl == null) {
+            baseUrl = "https://api." + accessKey.getDomain() + "/repository";
+        } else if (baseUrl.endsWith("/")) {
+            baseUrl.substring(0, baseUrl.length() - 1);
```
jason-chan-lf: The `substring` function returns the modified string, I don't think it updates baseUrl. So you still need to assign it.
https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#substring(int,%20int)

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
lf-jason-jia: Hi @jason-chan-lf , it has been changed to non-static.

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
steven-xu-lf: ok done

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
lf-jason-jia: @sxu1998 I think we discussed this before, there's no need to make it static. Let's change it back.

src/test/java/integration/BaseTest.java
steven-xu-lf: ok done

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
steven-xu-lf: ok done

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
steven-xu-lf: @lf-jason-jia ok so now that i removed the `static` keyword, is this comment now resolved?

lf-jason-jia: Yeah. I resolved it.

src/main/java/com/laserfiche/repository/api/SelfHostedInterceptor.java
```diff
@@ -0,0 +1,29 @@
+package com.laserfiche.repository.api;
+
+import com.laserfiche.api.client.apiserver.TokenClient;
+import com.laserfiche.api.client.httphandlers.*;
+import com.laserfiche.api.client.model.AccessKey;
```
lf-jason-jia: This import statement isn't needed.

steven-xu-lf: ok just reformatted that document

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -48,13 +49,23 @@ protected RepositoryApiClientImpl(String servicePrincipalKey, AccessKey accessKe
...
+        return new RepositoryApiClientImpl(interceptor, baseUrlDebug);
+    }
+
+    public static RepositoryApiClient createFromAccessKey(String servicePrincipalKey, AccessKey accessKey) {
+        return createFromAccessKey(servicePrincipalKey, accessKey, null);
     }
 
-    public static RepositoryApiClient CreateFromAccessKey(String servicePrincipalKey, AccessKey accessKey) {
-        return CreateFromAccessKey(servicePrincipalKey, accessKey, null);
+    public static RepositoryApiClient createFromUsernamePassword(String repoId, String username, String password,
```
jason-chan-lf: better to not abbreviate `repoId` -> `repositoryId`

jason-chan-lf (commented):
sorry, last comment. Also, remember to run the tests to check things are good

jason-chan-lf (approved):

PR #63, opened 2022-11-02
Use a separate call to queryString for the query parameters that thei…

samadpaydar3:
The generated code uses a single call to queryString method to set the query parameters in the url
There are two versions of this method in [kong.unirest.Path](https://github.com/Kong/unirest-java/blob/main/unirest/src/main/java/kong/unirest/Path.java):

1. one that takes an argument of type **Map<String, Object>**
2. one that takes two arguments (queryParameterName, and a collection of its values).

Now, the problem is that the code-generator generates a single call to queryString method, which is actually calling the first version of the method, and this method cannot handle query parameters that their value is an array. For such query parameters, it is needed to use the second version of the queryString method.

Our getEntryListing api has a query parameter named **fields**, which its value is a string array.

So, **the fix is to** update the code-generator to generate a separate call to queryString method, for each query parameter that its type is an array. For all the rest of the query parameters, a single call to the queryString is enough.

A new integration test is added to use getEntryListing api with fields, as all the previous tests use null for this parameter.

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/clients/impl/ApiClient.java
+++ b/src/main/java/com/laserfiche/repository/api/clients/impl/ApiClient.java
@@ -13,7 +13,9 @@ import kong.unirest.Headers;
 import kong.unirest.UnirestInstance;
 import org.threeten.bp.OffsetDateTime;
 
+import java.util.ArrayList;
 import java.util.HashMap;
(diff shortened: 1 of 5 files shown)


lf-jason-jia (approved):
Thanks for finding that API!

PR #69, opened 2022-11-08
Offband apiserver

lf-jason-jia:
1. Several changes to the accessors that haven't been included (eg. boolean getIsXxx is now changed to isXxx).
2. Include newly generated code that's due to recent APIServer offband release.
3. Bump version number to 1.1.0 (from 1.0.2) as we have added self-hosted support.

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/clients/impl/model/FuzzyType.java
+++ b/src/main/java/com/laserfiche/repository/api/clients/impl/model/FuzzyType.java
@@ -5,18 +5,18 @@ import com.fasterxml.jackson.annotation.JsonValue;
 
 public enum FuzzyType {
 
-    NUMBER_0(0),
-    NUMBER_1(1),
-    NUMBER_2(2);
+    NONE("None"),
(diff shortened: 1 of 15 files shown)


src/main/java/com/laserfiche/repository/api/clients/impl/model/FuzzyType.java
```diff
@@ -5,18 +5,18 @@
 
 public enum FuzzyType {
 
-    NUMBER_0(0),
-    NUMBER_1(1),
-    NUMBER_2(2);
+    NONE("None"),
+    PERCENTAGE("Percentage"),
+    NUMBEROFLETTERS("NumberOfLetters");
```
samadpaydar3: Isn't it better to have this renamed to NUMBER_OF_LETTERS to make it more readable?

lf-jason-jia: I updated the code. Thanks!

samadpaydar3 (approved):

PR #72, opened 2022-11-11
Fix bug by handling 409 status in ImportDocument in the success path instead of error path.

samadpaydar3:
For importDocument api, if we have partial success, APIServer does not return a success status code (e.g. 207). Instead, it returns 409. As in this case the response includes an CreateEntryResult object, we should handle this status code like other success status code, not like error status codes, (i.e. we need to return the CreateEntryResult object, instead of throwing exception)

So, to fix the issue, the code generator is updated to (only for importDocument api) handle 409 in the success path, instead of error path.

A new integration test is added for this fix: importDocument_Returns_Detail_When_PartialSuccess_Happens(
Some minor changes:

- Reduce duplication by initializing body once, instead of twice (one in the success and one in the error path)
- Some minor formatting changes
- Adding an enum for ErorrSource, to be used in the integration tests
- Removing unnecessary **throws** expression from some integration tests
- Removing some unneeded/duplicate assertions

### DIFF
--- a/src/test/java/integration/BaseTest.java
+++ b/src/test/java/integration/BaseTest.java
@@ -36,6 +36,9 @@ public class BaseTest {
     public static void setUp() {
         spKey = System.getenv("SERVICE_PRINCIPAL_KEY");
         String accessKeyBase64 = System.getenv("ACCESS_KEY");
+        if (!nullOrEmpty(accessKeyBase64)) {
+            accessKey = AccessKey.createFromBase64EncodedAccessKey(accessKeyBase64);
+        }
         repoId = System.getenv("REPOSITORY_ID");
         username = System.getenv("APISERVER_USERNAME");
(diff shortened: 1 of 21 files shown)


src/test/java/integration/BaseTest.java
```diff
@@ -36,6 +36,9 @@ public class BaseTest {
     public static void setUp() {
         spKey = System.getenv("SERVICE_PRINCIPAL_KEY");
         String accessKeyBase64 = System.getenv("ACCESS_KEY");
+        if (!nullOrEmpty(accessKeyBase64)) {
```
samadpaydar3, line 39: before this, if config is loaded from environment (instead of .env file), then accessKey was not initialized, causing the test not being executed.

lf-jason-jia (approved):

PR #76, opened 2022-12-01
Update enum classes for better naming

lf-jason-jia:
The enum constants from SwaggerCodegen doesn't contain _ to delimit words, which hurts readability. In this PR, we fixed that.


jason-chan-lf:
should HitType enum also be updated?
https://github.com/Laserfiche/lf-repository-api-client-java/blob/1.x/src/main/java/com/laserfiche/repository/api/clients/impl/model/HitType.java

samadpaydar3 (approved):

PR #78, opened 2022-12-02
Story/404922 encapsulate params

samadpaydar3:
The client implementations, e.g. EntriesClient, include methods with several parameters, while not all of those parameters are required. As Java does not support default value for method parameters, we decided to refactor the code generator to encapsulate the parameters of such methods in a separate class, and use that class as the parameter type. This way, when an api method is being called, it is not necessary to pass null arguments to the method.

In this PR:

- for each api method, a new class is created to encapsulate its parameters.
- that class is used as the parameter type for that api method
- As now the client doesn't need to send arguments for all the method parameters, there is no need to make the parameters nullable. So, the parameters are changed from wrapper classes to the primitive data types. For instance, from Integer to int.
Before this, we used null-check to decide which arguments need to be ignored, when using method parameters in the query parameters or in path.... This is changed as not the parameters are not Wrapper class, but primitive data type. So, the code ignores the arguments which are equal to the default value of the parameter. This means the method getNonNullParameters in the client library is replaced with getParametersWithNonDefaultValue.
- Integration tests are refactored to use the encapsulation classes

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/clients/params/ParametersForAssignEntryLinks.java
+++ b/src/main/java/com/laserfiche/repository/api/clients/params/ParametersForAssignEntryLinks.java
@@ -1,24 +1,22 @@
 package com.laserfiche.repository.api.clients.params;
 
-import com.laserfiche.repository.api.clients.impl.model.*;
-import java.io.InputStream;
+import com.laserfiche.repository.api.clients.impl.model.PutLinksRequest;
+
 import java.util.List;
-import java.util.Map;
-import java.util.function.Consumer;
 
 public class ParametersForAssignEntryLinks {
 
     /**
      * The request repository ID.
      */
-    String repoId;
+    private String repoId;
 
     /**
      * The requested entry ID.
      */
-    int entryId;
+    private int entryId;
 
-    List<PutLinksRequest> requestBody;
+    private List<PutLinksRequest> requestBody;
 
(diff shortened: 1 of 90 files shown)


lf-jason-jia (approved):

src/main/java/com/laserfiche/repository/api/clients/params/ParametersForAssignEntryLinks.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * The request repository ID.
+     */
+    private String repoId;
+
+    /**
+     * The requested entry ID.
+     */
+    private int entryId;
```
jason-chan-lf, line 17: The documentation is added on a private property, so users won't be able to see this documentation? Should documentation be added on the getters like in the model classes?
https://github.com/Laserfiche/lf-repository-api-client-java/blob/3a648e9f15ed9d1904fe70c35c650f2aa76b072f/src/main/java/com/laserfiche/repository/api/clients/impl/model/ContextHit.java#L120-L123

samadpaydar3: Sorry Jason,
I noticed your comment after I merge in the pr.

Yes. I will update the pr.

samadpaydar3: Updated.

PR #81, opened 2022-12-07
refactor baseTest file

steven-xu-lf:
- added a helper `getEnvironmentVariable` function in the basetest file
- removed abbreviated variable names
- refactored basetest implementation to be the same as the sample project implementation
- did not update the changelog since this is only involving the integration tests
- added the `IS_NOT_GITHUB_ENVIRONMENT` variable so the integration tests can also run on github when pipeline is enabled

### DIFF
--- a/src/test/java/integration/BaseTest.java
+++ b/src/test/java/integration/BaseTest.java
@@ -57,9 +57,6 @@ public class BaseTest {
                     getEnvironmentVariable(AUTHORIZATION_TYPE));
         }
         repositoryId = getEnvironmentVariable(REPOSITORY_ID);
-        if (nullOrEmpty(repositoryId)) {
-            throw new IllegalStateException("Environment variable REPOSITORY_ID does not exist.");
-        }
         if (authorizationType == AuthorizationType.CLOUD_ACCESS_KEY) {
             servicePrincipalKey = getEnvironmentVariable(SERVICE_PRINCIPAL_KEY);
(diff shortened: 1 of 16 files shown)


src/test/java/integration/BaseTest.java
```diff
@@ -22,78 +22,86 @@ enum AuthorizationType {
...
+    protected static String servicePrincipalKey;
     protected static AccessKey accessKey;
-    protected static String repoId;
+    protected static String repositoryId;
     protected static Map<String, String> testHeaders;
     protected static RepositoryApiClient repositoryApiClient;
+    protected static String testHeaderValue;
     protected static String username;
     protected static String password;
     protected static String baseUrl;
```
lf-jason-jia: Does it make sense to make the rest of field `private static final`?

steven-xu-lf: yea so i think it probably makes more sense to make the rest of the fields `private static` other than the `repositoryId` and `repositoryApiClient` and since the remaining variables cannot be initialized right away and need to be assigned at certain parts of the implementation. Both the `repositoryId` and `repositoryApiClient` variables are used by the subclasses which is why i prefer to keep it at protected

lf-jason-jia: Got it. Thanks for the explanation.

lf-jason-jia (approved):

PR #82, opened 2022-12-07
Share classes with client core library

lf-jason-jia:
In this PR, we removed classes that exist both or are functionally equivalent in this project and the client core library (a dependency of this project). The idea is, the client core library (as its name suggests) should be a place where the core classes and methods are hosted.

So, removed classes are:

- ApiException: this is a request from a customer, where they have to awkwardly catch two ApiExceptions because one is from the core library, the other is from this library. After this PR, they no longer need to do this.
- ProblemDetails: this class also exists in the core library, no need to have our own.
- OffsetDateTimeDeserializer: the ObjectMapper implementation of the client core library also has this identical class so we don't have to host it.
- RepositoryClientObjectMapper.java: this class is functionally equivalent to the ObjectMapper implementation in the core library.

In ApiClient class, we had a Jackson ObjectMapper. However, we already have an ObjectMapper implementation in client core. So, part of this PR also replaces it with the client core one. As a result, each client will need an instance of ObjectMapper passed in from the constructor (so all clients share the same HTTP client and ObjectMapper, as opposed to shared HTTP client and separated ObjectMapper). As a result, the ObjectMapper implementation doesn't throw a `JsonProcessingException`, instead, just `RuntimeException`, so we updated the try-catch blocks too.


src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -31,27 +32,30 @@ protected RepositoryApiClientImpl(RepositoryApiClientInterceptor interceptor, St
...
-        fieldDefinitionsClient = new FieldDefinitionsClientImpl(baseUrl, httpClient);
-        linkDefinitionsClient = new LinkDefinitionsClientImpl(baseUrl, httpClient);
-        repositoriesClient = new RepositoriesClientImpl(baseUrl, httpClient);
-        searchesClient = new SearchesClientImpl(baseUrl, httpClient);
-        serverSessionClient = new ServerSessionClientImpl(baseUrl, httpClient);
-        simpleSearchesClient = new SimpleSearchesClientImpl(baseUrl, httpClient);
-        tagDefinitionsClient = new TagDefinitionsClientImpl(baseUrl, httpClient);
-        tasksClient = new TasksClientImpl(baseUrl, httpClient);
-        templateDefinitionsClient = new TemplateDefinitionsClientImpl(baseUrl, httpClient);
+        attributesClient = new AttributesClientImpl(baseUrl, httpClient, objectMapper);
```
samadpaydar3: I think the ...ClientImpl's constructor can retrieve the object mapper from httpClient, like
`httpClient.config().getObjectMapper()
`If yes, we can remove the third parameter

lf-jason-jia: Good point. I will update the code here and for the code generator.

lf-jason-jia: Good morning. Sorry for the delay. I updated the code (both here and the code generator) to remove the parameter from constructor and get the ObjectMapper from the HTTP client.

samadpaydar3 (approved):

PR #84, opened 2022-12-20
Update javadoc

jason-chan-lf:
- Add missing javadoc
- Add javadoc links to classes and methods
- Add javadoc for setters for encapsulated parameter classes because users will be using the setters more often and better to call out that it uses the builder design pattern.
- Add missing encapsulated parameter javadoc in ForEach methods
- Disable generating javadoc in implementation classes. ForEach and NextLink methods already had their javadoc disabled in implementations
- Fix error in using closed repository api client when running all integration tests 
- Fix test assertion strings


src/main/java/com/laserfiche/repository/api/clients/params/ParametersForAssignEntryLinks.java
```diff
@@ -40,7 +55,7 @@ public ParametersForAssignEntryLinks setEntryId(int entryId) {
     /**
      * The requested entry ID.
      *
-     * @return int The return value
+     * @return {@link int} The return value
```
samadpaydar3: I think link does not work for primitive data types, like int, float.
When I try to ctrl+click on it, it is not clickable.

src/main/java/com/laserfiche/repository/api/clients/params/ParametersForAssignFieldValues.java
```diff
@@ -4,6 +4,9 @@
 
 import java.util.Map;
 
+/**
+ * The encapsulated parameters for {@link com.laserfiche.repository.api.clients.impl.EntriesClientImpl#assignFieldValues(ParametersForAssignFieldValues) assignFieldValues}.
```
samadpaydar3, line 8: This is not clear for me.
Shouldn't it be just 
{@link com.laserfiche.repository.api.clients.impl.EntriesClientImpl#assignFieldValues(ParametersForAssignFieldValues)}


instead of 
{@link com.laserfiche.repository.api.clients.impl.EntriesClientImpl#assignFieldValues(ParametersForAssignFieldValues) assignFieldValues}

src/main/java/com/laserfiche/repository/api/clients/params/ParametersForCancelOperation.java
```diff
@@ -12,6 +15,12 @@ public class ParametersForCancelOperation {
      */
     private String operationToken;
 
+    /**
+     * Builder for setting the repoId parameter.
```
samadpaydar3: I am not sure if it is good to mention Builder here, as we are not actually following builder pattern. I mean, when using Builder pattern, something like this is expected 
new ParametersForCancelOperationBuilder().setter1().setter2().setter3()....build();
but we provide no build method, and the client needs to explicitly call the constructor, like
new ParametersForCancelOperation().setter1().setter2()...

src/main/java/com/laserfiche/repository/api/clients/params/ParametersForCancelOrCloseSearch.java
```diff
@@ -12,6 +15,12 @@ public class ParametersForCancelOrCloseSearch {
      */
     private String searchToken;
 
+    /**
+     * Builder for setting the repoId parameter.
```
samadpaydar3: maybe better to use "attribute" instead of "parameter", i.e.
Builder for setting the repoId attribute.

jason-chan-lf: I used the term "parameter" to follow the class name since it is like "ParametersFor..."

samadpaydar3: Correct. I got it.

src/main/java/com/laserfiche/repository/api/clients/params/ParametersForAssignFieldValues.java
jason-chan-lf: If the javadoc was `{@link com.laserfiche.repository.api.clients.impl.EntriesClientImpl#assignFieldValues(ParametersForAssignFieldValues)}` then it would look like this in the ide.
image

If the javadoc was `{@link com.laserfiche.repository.api.clients.impl.EntriesClientImpl#assignFieldValues(ParametersForAssignFieldValues) assignFieldValues}` then it would look like this in the ide.
image

The second parameter in the `@link` is a text label to make the documentation less bloated with the namespace information. But if you think that information is good, then I can use the first option instead. Or maybe I can change the text label to `EntriesClientImpl.assignFieldValues` if you think that is better

samadpaydar3: Yes. You are correct. The second screenshot is better, as it is less verbose. Also the assignFieldValues is clickable, so, don't think it needs to be the fully-qualified name.
So, the current code is ok.

src/main/java/com/laserfiche/repository/api/clients/params/ParametersForCancelOperation.java
jason-chan-lf: Update the description to use the following format
`Sets the value of the __ parameter and returns the current object, to enable chaining further setters.`

src/main/java/com/laserfiche/repository/api/clients/params/ParametersForAssignEntryLinks.java
jason-chan-lf: I removed the `@link` for primitive types. thanks for catching it.

samadpaydar3 (approved):

PR #86, opened 2023-01-09
add afterSend() implementation and refactored the majority of the existing implementation

steven-xu-lf:
Added `afterSend()` and retry implementation as `sendRequestParseResponse()` method in the `ApiClient` class which includes
  - 401 status code check to call the `afterSend()` function
  - `isIdempotent()` check to call the `afterSend()` function 
  - Wrapped a while loop around each API implementation for retry logic and call `beforeSend()` in the beginning of the while loop  
  - Refactoring almost all API methods to call `sendRequestParseResponse()`

Refactored implementation for `importDocument` edge case
Refactored the `beforeSend()` method to the `exportDocument` API edge case (the `afterSend()` method will be done in another PR)

Refactored the entire structure of almost all of the implementation by:
- Adding a common sendAsync type of method `sendRequestParseResponse(
            UnirestInstance httpClient,
            ObjectMapper objectMapper, Class<TResponse> deserializedResponseType,
            HttpRequestHandler httpRequestHandler, String url, String requestMethod,
            String contentType,
            Object requestBody,
            String queryStringFields,
            Collection<?> queryStringFieldList,
            Map<String, Object> queryParameters, Map<String, Object> pathParameters,
            Map<String, String> headerParametersWithStringTypeValue,
            boolean isDynamicFieldValues)` method in the `ApiClient` class for almost all APIs to call
- removed both Self Hosted and OAuth `interceptors` and added a `beforeSend()` method in the `ApiClient` class which will call the `beforeSend()` method in the client core repo and modify the request URL
- Added additional helper functions in the `ApiClient` class such as: `isRetryableStatusCode()`, `combineURLs`, `isJsonResponse`, `createApiException`, `shouldThrowException`
- Added the `private HttpRequestHandler httpRequestHandler;` field to each `clientImpl` class and passed it in as a parameter from the `RepositoryApiClientImpl` class

Added a minor assertion check to an `AttributeClient` integration test 

Ran all the integration tests against OAuth and Selfhosted


src/main/java/com/laserfiche/repository/api/clients/impl/ApiClient.java
```diff
@@ -27,15 +31,16 @@ public ApiClient(String baseUrl, UnirestInstance httpClient) {
                 .getObjectMapper();
     }
 
-    protected String mergeMaxSizeIntoPrefer(int maxSize, String prefer) {
+    protected static String mergeMaxSizeIntoPrefer(int maxSize, String prefer) {
```
steven-xu-lf: move all the static functions in a utility new class so they do not need to be generated

src/main/java/com/laserfiche/repository/api/clients/impl/ApiClient.java
```diff
@@ -88,10 +93,250 @@ protected String toJson(Object object) {
...
+        if (err
+                .getProblemDetails()
+                .getTitle()
+                .equals("The user account name or password is incorrect. [9010]")) {
+            return true;
+        }
+        return false;
+    }
+
+    protected static <TResponse> TResponse sendRequestParseResponse(
```
steven-xu-lf: consider using lambda functions to be passed in instead of all these parameters due to it being easier to check status codes

src/main/java/com/laserfiche/repository/api/clients/impl/ApiClient.java
steven-xu-lf: moved all static functions to the `ApiUtilClient` class except for `sendRequestWithRetry` and the new created `parseDictionaryResponse()` function since they are more closely related to the ApiClient and could be frequently changed in the future

src/main/java/com/laserfiche/repository/api/clients/impl/ApiClient.java
steven-xu-lf: added a lambda parse function for every API client except for 2 additional edge cases (i.e. `getDocumentContentType` and `getDynamicFieldValues`)
For those 2 edge cases:
added a lambda function as a helper function in the `ApiClient` class for the `getDynamicFields` API
removed lambda function for getDocumentType since it was already included in the `sendResponse` function

steven-xu-lf (commented):
**Note:** I noticed that the lambda function already handles all the json response checking so i just removed that part of the implementation in the sendResponse 

Note #2: Removed the `shouldThrowException()` method since I think all the APIServerException in this case should be thrown 

Note #3: Removed all retrying error message print statements since it is not necessary to send that message to the user

jason-chan-lf (approved):

PR #88, opened 2023-01-17
Return CreateEntryResult for import doc api with an invalid parent ID

steven-xu-lf:
- Return `CreateEntryResult` type of response for an import document request with an invalid parent ID
- Added 2 additional integration tests
- Updated changelog
- Updated pom.xml file
- Ensured that the other import doc test cases were still running

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
+++ b/src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
@@ -204,7 +204,7 @@ public class EntriesClientImpl extends ApiClient implements EntriesClient {
                         if (response.getOperations() == null) {
                             throw new IllegalStateException();
                         }
-                        return objectMapper.readValue(jsonString, CreateEntryResult.class);
+                        return response;
                     } catch (JsonProcessingException | IllegalStateException e) {
                         Map<String, String> headersMap = getHeadersMap(httpResponse.getHeaders());
                         ProblemDetails problemDetails;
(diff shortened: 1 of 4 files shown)


src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
```diff
@@ -198,21 +198,19 @@ public CompletableFuture<CreateEntryResult> importDocument(String repoId, Intege
...
-                            return null;
+                    String jsonString = new JSONObject(body).toString();
+                    try {
+                        CreateEntryResult response = objectMapper.readValue(jsonString, CreateEntryResult.class);
+                        if (response.getOperations() == null) {
+                            throw new IllegalStateException();
                         }
-                    } else {
-                        ProblemDetails problemDetails;
+                        return objectMapper.readValue(jsonString, CreateEntryResult.class);
```
jason-chan-lf: could return the `response` instead of calling `objectMapper.readValue` again

src/test/java/integration/ImportDocumentApiTest.java
```diff
@@ -248,4 +248,49 @@ void importDocument_Returns_Detail_When_PartialSuccess_Happens() {
...
+                .getEntryCreate()
+                .getExceptions()
+                .get(0);
+        assertTrue(entryCreateException
+                .getMessage()
+                .startsWith("Entry not found."));
+        assertEquals(HttpURLConnection.HTTP_NOT_FOUND, entryCreateException.getStatusCode());
+        assertEquals(ErrorSource.LASERFICHE_SERVER.getName(), entryCreateException.getErrorSource());
+    }
+
```
jason-chan-lf, line 297: could you also add a test that causes a ProblemDetails response to be returned? I saw none of the current import api tests test this case yet. It would be nice to know that the import api can handle this response type.

steven-xu-lf: ok added a test case with an invalid repo as a part of the request

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
steven-xu-lf: ok done

jason-chan-lf (approved):

PR #89, opened 2023-01-23
added no jwt integration test

steven-xu-lf:
- added no jwt self hosted integration test
- updated baseTest to account for no jwt self hosted


steven-xu-lf:
created a new PR for the same issue

PR #90, opened 2023-01-24
Update api error handling

### LINKED ISSUE #87: Improve error handling
- errors thrown by the client library should be consistent (ie try to throw ApiException) so that clients can know how to handle the errors

jason-chan-lf:
Resolves https://github.com/Laserfiche/lf-repository-api-client-java/issues/87
- Error handling has been updated to
  - throw ApiException when unable to deserialize response
  - throw ApiException when an undocumented status code is returned
- EntriesClient.importDocument has been updated to only return CreateEntryResult when a 200 status code is returned. When other error status codes return a CreateEntryResult, it will be a property in the ApiException.ProblemDetails.Extensions and the error will be thrown. 
- EntriesClient.getDocumentContentType error handling has been simplified

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/clients/impl/ApiClientUtils.java
+++ b/src/main/java/com/laserfiche/repository/api/clients/impl/ApiClientUtils.java
@@ -27,10 +27,15 @@ class ApiClientUtils {
         Collection<String> messages = new ArrayList<>();
         if (createEntryResult != null && createEntryResult.getOperations() != null) {
             CreateEntryOperations operations = createEntryResult.getOperations();
-            if (operations.getEntryCreate() != null)
+            if (operations.getEntryCreate() != null) {
+                Integer entryId = operations.getEntryCreate().getEntryId();
+                if (entryId != null && entryId > 0) {
+                    messages.add(String.format("EntryId=%s.", entryId));
+                }
                 messages.add(getErrorMessagesFromAPIServerExceptions(operations
                         .getEntryCreate()
                         .getExceptions()));
+            }
             if (operations.getSetEdoc() != null)
                 messages.add(getErrorMessagesFromAPIServerExceptions(operations
                         .getSetEdoc()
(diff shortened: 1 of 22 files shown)


src/main/java/com/laserfiche/repository/api/clients/impl/ApiClientUtils.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+/**
+ * Helper class containing utility functions for the RepositoryApiClient.
+ */
+class ApiClientUtils {
```
jason-chan-lf: make unit tests for this class

jason-chan-lf: added unit tests

src/main/java/com/laserfiche/repository/api/clients/impl/ApiClientUtils.java
```diff
@@ -10,9 +10,9 @@
 import java.util.stream.Collectors;
 
 /**
- * Helper class containing utility functions for the RepositoryApiClient.
+ * Internal helper class containing utility functions for the RepositoryApiClient.
```
steven-xu-lf, line 13: does the code generator also need to be updated after this change?

jason-chan-lf: This class isn't auto generated, since it isn't part of the swagger schema. I just updated the code generator to call this class when using the import function

steven-xu-lf: ok thanks for the clarification

steven-xu-lf (dismissed):

steven-xu-lf (approved):

PR #92, opened 2023-02-27
added exportDoc retry implementation

steven-xu-lf:
- added the afterSend retry logic for both exportDocument APIs
- updated 413 status code for the exportDocWithAuditReason API

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
+++ b/src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
@@ -600,6 +600,7 @@ public class EntriesClientImpl extends ApiClient implements EntriesClient {
             final RuntimeException[] exception = {null};
             int retryCount = 0;
             int maxRetries = 1;
+            final int[] statusCode = {0};
             boolean shouldRetry = true;
             while (retryCount <= maxRetries && shouldRetry) {
                 try {
@@ -613,6 +614,7 @@ public class EntriesClientImpl extends ApiClient implements EntriesClient {
                             .contentType("application/json")
                             .body(parameters.getRequestBody())
                             .thenConsume(rawResponse -> {
+                                statusCode[0] = rawResponse.getStatus();
                                 if (rawResponse.getStatus() == 200 || rawResponse.getStatus() == 206) {
                                     parameters
                                             .getInputStreamConsumer()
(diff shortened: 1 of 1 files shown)


src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
```diff
@@ -651,15 +656,20 @@ else if (rawResponse.getStatus() == 429)
                                         exception[0] = new RuntimeException(rawResponse.getStatusText());
```
jason-chan-lf: Just noticed this exception was missed to throw ApiException. Could you update this else exception to be something like `ApiException.create(rawResponse.getStatus(), headersMap, problemDetails, null);`

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
```diff
@@ -651,15 +656,20 @@ else if (rawResponse.getStatus() == 429)
...
+                        exception[0] = null;
                     }
                 } catch (Exception err) {
                     if (err instanceof ApiException) {
                         if (exception[0] != null) {
                             throw exception[0];
-                        } else {
-                            System.err.println(err);
                         }
                         break;
```
jason-chan-lf: a bit confused about the catch block. If the `err` is `ApiException` type but `exception[0]` is null then it will break the while loop and not throw any exception? Could this part be the same as the catch block in `ApiClientUtils.sendRequestWithRetry`? The new code is already throwing `exception[0]` if not retrying so do we still need the check for `exception[0] != null` in the catch?
```
            } catch (Exception err) {
                if (err instanceof ApiException || retryCount >= maxRetries) {
                    throw err;
                }
                shouldRetry = true;
            }
```

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
steven-xu-lf: ok in this case then guess i can just replace lines 634 - 656 with `exception[0] = ApiException.create(rawResponse.getStatus(), headersMap, problemDetails, null);` for both exportDocument APIs since it seems to be the same thing logically

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
steven-xu-lf: yea i agree, going with the catch block logic in the `ApiClientUtils.sendRequestWithRetry` function would probably be the way to go, i think this might be something that i forgot to update (not too sure why i had it there initially)

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
steven-xu-lf: ok done

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
steven-xu-lf: ok done

jason-chan-lf (approved):

PR #95, opened 2023-03-03
replaced all primitive type variables

steven-xu-lf:
In the `com.laserfiche.repository.api.clients.params` package
- replaced `boolean` with `Boolean` in all classes
- replaced `int` with `Integer` in all classes


steven-xu-lf:
closing this issue because the primitive changes in v2 were intentional

steven-xu-lf:
re-opening this issue due new mob session proposal changes

jason-chan-lf (approved):

PR #96, opened 2023-03-06
fixed importDoc with an invalid repo error message bug

steven-xu-lf:
- changed the exception error message from `HTTP status code 404` to `Error: Repository with the given Id not found or no connection could be made.` for sending an importDocument API with an invalid repo query parameter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@
   - 401 status code response is being returned
   - Idempotent http request
 
+### Fixes
+- For sending an importDocument API with an invalid repo, changed error message displayed from `HTTP status code 404` to `Error: Repository with the given Id not found or no connection could be made.`
+
 ## 1.2.0
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,6 @@
   - 401 status code response is being returned
   - Idempotent http request
 
-### Fixes
-- For sending an importDocument API with an invalid repo, changed error message displayed from `HTTP status code 404` to `Error: Repository with the given Id not found or no connection could be made.`
-
 ## 1.2.0
 
 ### Features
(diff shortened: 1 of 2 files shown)


src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
```diff
@@ -170,18 +170,29 @@ public CreateEntryResult importDocument(ParametersForImportDocument parameters)
...
-                                    if (result != null && result.getOperations() != null) {
-                                        problemDetails
-                                                .getExtensions()
-                                                .put(CreateEntryResult.class.getSimpleName(), result);
-                                        String summary = ApiClientUtils.getCreateEntryResultSummary(result);
-                                        if (summary != null && !summary
-                                                .trim()
-                                                .isEmpty())
-                                            problemDetails.setTitle(summary);
+                                    if (httpResponse
```
jason-chan-lf, line 173: maybe it would be better to check if the content-type header has `application/problem+json` to determine if the response is problem details

CHANGELOG.md
```diff
@@ -25,6 +25,9 @@
   - 401 status code response is being returned
   - Idempotent http request
 
+### Fixes
+- For sending an importDocument API with an invalid repo, changed error message displayed from `HTTP status code 404` to `Error: Repository with the given Id not found or no connection could be made.`
```
jason-chan-lf: since v2 isn't released yet, I don't think this needs to be mentioned as a fix. We already mentioned above that `EntriesClient.importDocument` has a breaking change when handling errors

src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
steven-xu-lf: ok done

CHANGELOG.md
steven-xu-lf: ok removed

jason-chan-lf (approved):

PR #97, opened 2023-03-08
added initial implementation prior to code generator changes

steven-xu-lf:
Implemented two new functions for export that return a stream. The user is responsible for closing the stream.
- InputStream exportDocumentAsStream(ParametersForExportDocument parameters);
- InputStream exportDocumentWithAuditReasonAsStream(ParametersForExportDocumentWithAuditReason parameters);
- Add integration tests for both methods
- Add integration tests for the void exportDocumentWithAuditReason(ParametersForExportDocumentWithAuditReason parameters); too (just noticed that they were missing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 2.0.0
+
 ### Features
+
 - **[BREAKING]** For every API, the parameters are now encapsulated into a parameter class.
 - **[BREAKING]** Enum values are now following Java naming convention.
 - **[BREAKING]** Every async API are replaced with normal blocking API.
@@ -8,6 +10,7 @@
   - 401 status code response is being returned
   - Idempotent http request
 - Added an option when constructing the `RepositoryApiClient` to specify the requested scope(s) for the access token.
+- Added both `exportDocumentAsStream` and `exportDocumentWithAuditReasonAsStream` API methods in the `EntriesClientImpl` object along with the `EntriesClient` interface
 
 ### Fixes
 - The ApiClient class is now abstract.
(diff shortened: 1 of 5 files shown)


src/main/java/com/laserfiche/repository/api/clients/impl/ApiClientUtils.java
```diff
@@ -89,6 +89,12 @@ protected static ApiException createApiException(HttpResponse<Object> httpRespon
         return ApiException.create(statusCode, headersMap, problemDetails, null);
     }
 
+    protected static ApiException createApiExceptionForExportDocumentAsStream(HttpResponse<byte[]> httpResponse, ProblemDetails problemDetails) {
```
jason-chan-lf: This function looks the same as the above `createApiException` with the only difference being httpResponse has type `HttpResponse<byte[]>`. Maybe instead you could update `createApiException` with parameter `HttpResponse<?>` so that it can be used by the `exportDocumentAsStream` function

CHANGELOG.md
```diff
@@ -28,6 +28,11 @@
   }
   ```
 
+- Added retry implementation when making an API call which retries when:
```
jason-chan-lf: This changelog for retry got moved above, so this can be removed

src/main/java/com/laserfiche/repository/api/clients/impl/ApiClientUtils.java
steven-xu-lf: ok done, thanks!

CHANGELOG.md
steven-xu-lf: ok done, and also moved the line about `exportDocumentAsStream` above the fixes section

jason-chan-lf (approved):

PR #100, opened 2023-03-24
update to maven central dependency, updated CHANGELOG.md, and updated…

steven-xu-lf:
- updated to maven central dependency
- updated changelog
- renamed pipeline name
- skipped/disabled all entry tests for now due to rate limiting issue that keeps popping up in the pipeline
- will test the publish to production step once this PR is approved


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.0.2
+
+### Fixes
+
+- Updated the [Java Client Core Repository](https://github.com/Laserfiche/lf-api-client-core-java) dependency in pom.xml file to use the maven central version of rather than the `SNAPSHOT` version
```
jason-chan-lf: Could simplify this to something like
Updated the [Java Client Core Repository](https://github.com/Laserfiche/lf-api-client-core-java) dependency to release version 2.2.0

CHANGELOG.md
```diff
@@ -12,6 +18,8 @@
 - Added an option when constructing the `RepositoryApiClient` to specify the requested scope(s) for the access token.
 - Added both `exportDocumentAsStream` and `exportDocumentWithAuditReasonAsStream` API methods in the `EntriesClientImpl` object along with the `EntriesClient` interface
 
+#### Note: accidently published version 2.0.1 so there is no difference between version 2.0.1 and 2.0.0
```
jason-chan-lf: Probably not necessary to mention this. I think if there are no changelog mentioned for a version, like 2.0.1, then it can be assumed there are no changes that impact users.

CHANGELOG.md
steven-xu-lf: ok done

CHANGELOG.md
steven-xu-lf: ok done

jason-chan-lf (approved):

PR #101, opened 2023-03-30
added getSelfHosted repo implementation and integration test

steven-xu-lf:
- added `getSelfHostedRepoList()` method for self hosted users to get all self hosted repos without any authorization
- added a conditional integration test case for it
- updated version and changelog
- refactor all `getRepoList` test cases
- avoided calling the `getRepositoryList()` since the before and aftersend logic will be applied if calling that we want to keep it as simple as possible 
- decided to not add an encapsulation class for new method (like all the `nextlink` functions) and just keep it simple

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/clients/impl/RepositoriesClientImpl.java
+++ b/src/main/java/com/laserfiche/repository/api/clients/impl/RepositoriesClientImpl.java
@@ -1,15 +1,12 @@
 package com.laserfiche.repository.api.clients.impl;
 
 import com.laserfiche.api.client.deserialization.ProblemDetailsDeserializer;
-import com.laserfiche.api.client.deserialization.TokenClientObjectMapper;
 import com.laserfiche.api.client.httphandlers.HttpRequestHandler;
 import com.laserfiche.api.client.model.ApiException;
 import com.laserfiche.api.client.model.ProblemDetails;
 import com.laserfiche.repository.api.clients.RepositoriesClient;
 import com.laserfiche.repository.api.clients.impl.model.RepositoryInfo;
 import kong.unirest.HttpResponse;
-import kong.unirest.ObjectMapper;
-import kong.unirest.Unirest;
 import kong.unirest.UnirestInstance;
 import kong.unirest.json.JSONArray;
 import kong.unirest.json.JSONObject;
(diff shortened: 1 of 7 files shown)


src/main/java/com/laserfiche/repository/api/clients/impl/RepositoriesClientImpl.java
```diff
@@ -48,6 +51,52 @@ public RepositoryInfo[] getRepositoryList() {
...
-        return ApiClientUtils.sendRequestWithRetry(httpClient, httpRequestHandler, baseUrl + "/v1/Repositories", "GET", null, null, null, null, null, null, new HashMap<String, String>(), false, parseResponse);
+        return ApiClientUtils.sendRequestWithRetry(httpClient, httpRequestHandler, baseUrl + "/v1/Repositories", "GET",
+                null, null, null, null, null, null, new HashMap<String, String>(), false, parseResponse);
+    }
+
+    /**
+     * Returns the repository resource list that current user has access to given the API server base URL. Only available in Laserfiche Self-Hosted.
+     *
+     * @param url API server base URL e.g., https://{APIServerName}/LFRepositoryAPI
+     * @return Get the respository resource list successfully.
```
jason-chan-lf: repository typo

src/main/java/com/laserfiche/repository/api/clients/impl/RepositoriesClientImpl.java
```diff
@@ -48,6 +51,52 @@ public RepositoryInfo[] getRepositoryList() {
...
+        Map<String, String> headerKeyValuePairs = new HashMap<>();
+        headerKeyValuePairs.put("accept", "application/json");
+        HttpResponse<Object> httpResponse = null;
+        String responseJson;
+        if (url.endsWith("/")) {
+            url = url.substring(0, url.length() - 1);
+        }
+        String baseUrl = url + "/v1/Repositories";
+        ObjectMapper objectMapper = new TokenClientObjectMapper();
+        UnirestInstance httpClient = Unirest.spawnInstance();
```
jason-chan-lf: should dispose the httpClient afterwards. Could use a try-with-resource pattern on the httpClient
https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

src/main/java/com/laserfiche/repository/api/clients/impl/RepositoriesClientImpl.java
```diff
@@ -48,6 +51,52 @@ public RepositoryInfo[] getRepositoryList() {
...
+                null, null, null, null, null, null, new HashMap<String, String>(), false, parseResponse);
+    }
+
+    /**
+     * Returns the repository resource list that current user has access to given the API server base URL. Only available in Laserfiche Self-Hosted.
+     *
+     * @param url API server base URL e.g., https://{APIServerName}/LFRepositoryAPI
+     * @return Get the respository resource list successfully.
+     */
+    public static RepositoryInfo[] getSelfHostedRepositoryList(String url) {
```
jason-chan-lf, line 64: should this follow the same pattern as the other api functions with a ParametersFor function argument?

steven-xu-lf: i just decided to avoid wrapping it in an encapsulation object and keep it simple like all the `nextlink` functions since it is only 1 parameter

src/main/java/com/laserfiche/repository/api/clients/impl/RepositoriesClientImpl.java
steven-xu-lf: yea that's good idea, ok done

src/main/java/com/laserfiche/repository/api/clients/impl/RepositoriesClientImpl.java
steven-xu-lf: ok done

jason-chan-lf (approved):

PR #102, opened 2023-04-18
added initial unit tests for apiclientUtil helper functions

steven-xu-lf:
Added unit test for some of the helper functions in the `ApiClientUtils` file, helper functions included:
- public functions (i.e. `isRetryableStatusCode`, `mergeMaxSizeIntoPrefer`, and `getHeadersMap`)
- functions containing parameters without any prior object instantiation 
- also added some non happy path unit tests 

Renamed existing `ApiClientUtilsTest` to `GetCreateEntryResultSumary`

### DIFF
--- a/src/test/java/unit/clients/impl/GetCreateEntryResultSumary.java
+++ b/src/test/java/unit/clients/impl/GetCreateEntryResultSumary.java
@@ -9,7 +9,7 @@ import java.util.List;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
-class ApiClientUtilsTest {
+class GetCreateEntryResultSumary {
     @Test
     void getCreateEntryResultSummary_NullCreateEntryResult() {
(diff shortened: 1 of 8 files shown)


src/test/java/unit/clients/impl/GetCreateEntryResultSumary.java
```diff
@@ -9,7 +9,7 @@
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
-class ApiClientUtilsTest {
```
jason-chan-lf: Can you put all these new unit tests in a folder called `ApiClientUtils` so that it is easy to tell when using the file explorer that these tests are for this class

steven-xu-lf: ok done

jason-chan-lf (approved):

PR #111, opened 2023-05-23
Add content type parameter when using importing document

jason-chan-lf:
- Fixed a bug in `EntriesClient.importDocument` where imported documents have an incorrect mime type. Clients will need to use the `ParametersForImportDocument.setContentType()` method to set the mime type of the imported document. 
- Added some missing java doc for Parameters classes


src/main/java/com/laserfiche/repository/api/clients/impl/EntriesClientImpl.java
```diff
@@ -146,7 +143,9 @@ public CreateEntryResult importDocument(ParametersForImportDocument parameters)
                             httpRequestHandler);
                     httpResponse = httpClient
                             .post(requestUrl)
-                            .field("electronicDocument", parameters.getInputStream(), parameters.getFileName())
+                            .field("electronicDocument", parameters.getInputStream(),
+                                    parameters.getContentType() == null ? ContentType.APPLICATION_OCTET_STREAM : ContentType.create(
```
jason-chan-lf, line 147: previously, no content type was set on the `electronicDocument` parameter. unirest will default to `ContentType.APPLICATION_OCTET_STREAM` in this case. Since we allow setting content type now, if content type is null, we will set it to the default `ContentType.APPLICATION_OCTET_STREAM` to keep similar behaviour as before.

lf-dfc (approved):

matt-yin-lf (approved):

PR #112, opened 2023-05-24
added gzip compression

steven-xu-lf:
- added gzip compression header when creating an http Client
- updated changelog

### DIFF
--- a/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
+++ b/src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
@@ -11,6 +11,7 @@ import kong.unirest.ObjectMapper;
 import kong.unirest.Unirest;
 import kong.unirest.UnirestInstance;
 
+import java.util.HashMap;
 import java.util.Map;
 
 /**
@@ -18,6 +19,7 @@ import java.util.Map;
  */
 public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseable {
     private Map<String, String> defaultHeaders;
+    private Map<String, String> gzipCompressionHeaders = new HashMap<>();
     private final UnirestInstance httpClient;
     private final AttributesClient attributesClient;
     private final AuditReasonsClient auditReasonsClient;
@@ -48,9 +50,8 @@ public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseab
 
         // Add compression header if a client is created
         if (httpClient != null) {
-            httpClient
-                    .config()
-                    .setDefaultHeader("Accept-Encoding", "gzip");
+            gzipCompressionHeaders.put("Accept-Encoding", "gzip");
+            setDefaultRequestHeaders(gzipCompressionHeaders);
         }
 
         // Initialize repository API clients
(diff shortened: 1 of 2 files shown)


src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -51,6 +46,13 @@ protected RepositoryApiClientImpl(String baseUrl, HttpRequestHandler httpHandler
                 .config()
                 .setObjectMapper(objectMapper);
 
+        // Add compression header if a client is created
+        if (httpClient != null) {
+            httpClient
+                    .config()
+                    .setDefaultHeader("Accept-Encoding", "gzip");
```
jason-chan-lf: If the user calls `getDefaultRequestHeaders`, I don't think this Accept-Encoding header will be returned. I think this should be fixed if you use the `setDefaultRequestHeaders` function to set the default header instead.

steven-xu-lf: ok done, and also updated the `setDefaultRequestHeaders()` function so that it won't automatically delete the gzip compression header when adding in another header

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -127,9 +130,14 @@ public static RepositoryApiClient createFromUsernamePassword(String repositoryId
     @Override
     public void setDefaultRequestHeaders(Map<String, String> defaultRequestHeaders) {
         defaultHeaders = defaultRequestHeaders;
-        httpClient
+        if (!httpClient
                 .config()
-                .clearDefaultHeaders();
+                .getDefaultHeaders()
+                .containsKey("Accept-Encoding")) {
```
jason-chan-lf: i'm not sure if `setDefaultRequestHeaders()` needs to be changed. If the user chooses to override the default request headers with new defaults, then we should let them clear the existing default headers

If the user wants to keep the existing defaults, then they can call `getDefaultRequestHeaders()` and add their new default headers to the returned Map and set it with `setDefaultRequestHeaders()`

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
```diff
@@ -23,6 +19,7 @@
  */
 public class RepositoryApiClientImpl implements RepositoryApiClient, AutoCloseable {
     private Map<String, String> defaultHeaders;
+    private Map<String, String> gzipCompressionHeaders = new HashMap<>();
```
jason-chan-lf: this doesn't need to be a class variable, could create it in the constructor instead.

steven-xu-lf: ok done

src/main/java/com/laserfiche/repository/api/RepositoryApiClientImpl.java
steven-xu-lf: ok reverted back

jason-chan-lf (approved):

PR #122, opened 2023-08-02
added test case conditions and conditional method

steven-xu-lf:
Added a conditional method to check if audit reasons exists
- added `doAuditReasonsNotExist()` conditional method which will return `true` if no audit reasons exists or when an exception gets thrown and return `false` if at least one audit reason exists
- if there are audit reasons, then run some certain test cases (i.e. all happy path `exportDocument_WithAuditReasons` test cases) and skip some other ones (i.e. all happy path `exportDocument` test cases)
- if there are no audit reasons, then run all happy path `exportDocument` test cases and skip all happy path `exportDocument_WithAuditReasons` test cases

### DIFF
--- a/src/test/java/com/laserfiche/repository/api/integration/ExportDocumentApiTest.java
+++ b/src/test/java/com/laserfiche/repository/api/integration/ExportDocumentApiTest.java
@@ -1,15 +1,21 @@
 package com.laserfiche.repository.api.integration;
 
-import static org.junit.jupiter.api.Assertions.*;
-
 import com.laserfiche.api.client.model.ApiException;
 import com.laserfiche.repository.api.clients.AuditReasonsClient;
 import com.laserfiche.repository.api.clients.EntriesClient;
 import com.laserfiche.repository.api.clients.impl.model.*;
-import com.laserfiche.repository.api.clients.params.*;
+import com.laserfiche.repository.api.clients.params.ParametersForExportDocument;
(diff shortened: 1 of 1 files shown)


lf-dfc:
> Added a conditional method to check if audit reasons exists
> 
> * added `doAuditReasonsNotExist()` conditional method which will return `true` if no audit reasons exists or when an exception gets thrown and return `false` if at least one audit reason exists
> * if there are audit reasons, then run some certain test cases (i.e. all happy path `exportDocument_WithAuditReasons` test cases) and skip some other ones (i.e. all happy path `exportDocument` test cases)
> * if there are no audit reasons, then run all happy path `exportDocument` test cases and skip all happy path `exportDocument_WithAuditReasons` test cases

Does this change also needed for other clients in other languages?

steven-xu-lf:
> > Added a conditional method to check if audit reasons exists
> > 
> > * added `doAuditReasonsNotExist()` conditional method which will return `true` if no audit reasons exists or when an exception gets thrown and return `false` if at least one audit reason exists
> > * if there are audit reasons, then run some certain test cases (i.e. all happy path `exportDocument_WithAuditReasons` test cases) and skip some other ones (i.e. all happy path `exportDocument` test cases)
> > * if there are no audit reasons, then run all happy path `exportDocument` test cases and skip all happy path `exportDocument_WithAuditReasons` test cases
> 
> Does this change also needed for other clients in other languages?
I don't think so since these test cases do not exist in the other 2 repo clients

lf-dfc (approved):

PR #126, opened 2023-10-03
Story/471858 generate client lib

samadpaydar3:
Create v2 Java client library
- Re-generated client library based on v2 swagger.json
- Added new integration tests, e.g. for large file import or for new features of Export/Import apis
- Renamed integration tests based on the changes in the name of the APIs they call
- For APIs, like listTasks, that take a `String[]` parameter, the setter method is updated to use Java VarArgs, i.e. use `String...` instead of `String[]` to make it easier to call the api with one value.


src/main/java/com/laserfiche/repository/api/clients/impl/model/FolderImportBody.java
```diff
@@ -0,0 +1,91 @@
...
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.v3.oas.annotations.media.Schema;
+
+import java.io.File;
+import java.util.Objects;
+
+@javax.annotation.Generated(value = "io.swagger.codegen.v3.generators.java.JavaClientCodegen")
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class FolderImportBody {
```
jason-chan-lf: I checked that this class isn't used anywhere. Does it need to be included?

samadpaydar3: Thank you for catching the issue. I removed the class.

jason-chan-lf (dismissed):

jason-chan-lf (approved):

PR #129, opened 2023-10-03
[gh-pages] Add v2 docs

jason-chan-lf:
- initial v2 documentation to be used to link in the migration guide


docs/v2/1.x/overview-summary.html
```diff
@@ -0,0 +1,160 @@
...
+  }
+  //-->
+</script>
+</div>
+<a name="skip.navbar.top">
+<!--   -->
+</a></div>
+<!-- ========= END OF TOP NAVBAR ========= -->
+<div class="header">
+<h1 class="title">Laserfiche Repository API Client V2 1.0.0 API</h1>
```
samadpaydar3: Isn't this second **API** redundant?

jason-chan-lf: Fixed. I had to update the pom.xml file when generating the javadoc. I'll submit a PR that will include this change later

samadpaydar3 (approved):

PR #150, opened 2025-02-11
update pom.xml and version

alexgomezlf:


pom.xml
shiyuanzhao-lf, line 1: does the jackson-version around Line 312 need to update?

shiyuanzhao-lf (approved):

